PR #4, opened 2018-01-05
Exposes endpoint to create upload-session 

AaronVasquez:
This is still a WIP


index.php
```diff
@@ -1,12 +1,44 @@
...
+/**
+ * HTTP Handlers
+ */
+
+$app = new \Slim\App;
+
+$app->post('/api/v1/upload-session', function (Request $request, Response $response, array $args) {
+    $json = json_decode($request->getBody(), true);
+
+    // TODO: Persist upload session data
```
AaronVasquez: @brunoblock @pandora23 
Will this be shared between nodes? Should I persist this in something like Redis?

pandora23:
We could probably stick with the DB we are thinking of MariaDB because it
does both relational tables and key/value stores and does in memory stuff
with persistence.

PR #18, opened 2018-01-11
Endpoint for uploading chunks

AaronVasquez:
- Creates datamap when a session is started
    - This is synchronous for now to avoid dealing with potential race conditions
- Implements /api/v1/upload-session endpoint to upload chunks

### DIFF
--- a/broker-node/app/DataMapping.php
+++ b/broker-node/app/DataMapping.php
@@ -22,90 +22,93 @@ class DataMapping extends Model
     protected $table = 'data_mappings';
 	protected $fillable = ['genesis_hash', 'hash', 'chunk_idx'];
     public $incrementing = false;  // UUID
-}
 
-/**
- * Builds and loads the datamap in DB given genesis hash and the number of file chunks
- *
- * @param string $genesis_hash A 64 char string
- * @param integer $file_chunk_count the number to generate
- * @return boolean $result response code
- */
-function buildMap($genesis_hash, $file_chunk_count) {
-    //we put the first chunk
-    $chunk_idx = 0;
-    App\DataMapping::create([
-        'genesis_hash' => $genesis_hash,
-        'hash' => $genesis_hash,
-        'chunk_idx' => $chunk_idx
-    ]);
+    /**
+     * Builds and loads the datamap in DB given genesis hash and the number of file chunks
+     *
+     * @param string $genesis_hash A 64 char string
+     * @param integer $file_chunk_count the number to generate
+     * @return boolean $result response code
+     */
+    public static function buildMap($genesis_hash, $file_chunk_count) {
+        //we put the first chunk
+        $chunk_idx = 0;
+        self::create([
+            'genesis_hash' => $genesis_hash,
+            'hash' => $genesis_hash,
+            'chunk_idx' => $chunk_idx
+        ]);
 
-    //it takes the ceiling then disregards any past the correct number
-    $hashes_per_db_update = 5;
-    $num_groups = ceil($file_chunk_count / $hashes_per_db_update);
+        //it takes the ceiling then disregards any past the correct number
+        $hashes_per_db_update = 5;
+        $num_groups = ceil(($file_chunk_count - 1) / $hashes_per_db_update);
 
-    //start process of getting hashes
-    $hash_generator = hashGenerator($genesis_hash,  $file_chunk_count);
+        //start process of getting hashes
+        $hash_generator = self::hashGenerator($genesis_hash,  $file_chunk_count);
 
-    for($i = 0; $i < $num_groups; $i++) {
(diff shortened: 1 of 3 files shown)


broker-node/app/Http/Controllers/UploadSessionController.php
```diff
@@ -77,7 +85,21 @@ public function edit($id)
...
+
+        $genesis_hash = $session['genesis_hash'];
+        $chunk = $request->input('chunk');
+
+        $data_mapping = DataMapping::where('genesis_hash', $genesis_hash)
+            ->where('chunk_idx', $chunk['idx'])
+            ->first();
+        if (empty($data_mapping)) return response('Datamap not found', 404);
+
+        // TODO: What to do with $data_mapping['hash']
```
AaronVasquez: @pandora23 Once I have the datamap, what needs to be done with the hash?

broker-node/app/DataMapping.php
```diff
@@ -22,90 +22,93 @@ public static function boot()
...
+     * @param string $genesis_hash A 64 char string
+     * @param integer $file_chunk_count the number to generate
+     * @return boolean $result response code
+     */
+    public static function buildMap($genesis_hash, $file_chunk_count) {
+        //we put the first chunk
+        $chunk_idx = 0;
+        self::create([
+            'genesis_hash' => $genesis_hash,
+            'hash' => $genesis_hash,
```
EdmundMai: shouldn't this be `'data' => $data`?

EdmundMai: ohhh you're saving the genesis hash as the first chunk

broker-node/app/Http/Controllers/UploadSessionController.php
```diff
@@ -2,6 +2,7 @@
 
 namespace App\Http\Controllers;
 
+use App\DataMapping;
```
EdmundMai: is this persistent?

broker-node/app/DataMapping.php
```diff
@@ -22,90 +22,93 @@ public static function boot()
...
+        //start process of getting hashes
+        $hash_generator = self::hashGenerator($genesis_hash,  $file_chunk_count);
+
+
+        for($i = 0; $i < $num_groups; $i++) {
+            $next_group = self::getNextNHashes($hash_generator, $hashes_per_db_update, $chunk_idx);
+            $chunk_idx += $hashes_per_db_update;
+
+            foreach($next_group as $chunk_idx_hash) {
+                [$curr_hash, $curr_chunk_idx] = $chunk_idx_hash;
```
EdmundMai: hm how does `$curr_hash` exist in this context? isn't this method called when we create the upload session (and not when we send the chunks)?

broker-node/app/Http/Controllers/UploadSessionController.php
EdmundMai: i think we need to make sure $data_mapping['hash'] == $chunk['hash'] to verify that they're coming from the same source?

broker-node/app/DataMapping.php
```diff
@@ -22,90 +22,93 @@ public static function boot()
...
+        //we put the first chunk
+        $chunk_idx = 0;
+        self::create([
+            'genesis_hash' => $genesis_hash,
+            'hash' => $genesis_hash,
+            'chunk_idx' => $chunk_idx
+        ]);
+
+        //it takes the ceiling then disregards any past the correct number
+        $hashes_per_db_update = 5;
```
EdmundMai: is this just to batch the DB transactions so that we don't hit the DB too many times?

broker-node/app/DataMapping.php
AaronVasquez: We don't save the chunk to the DB, it gets saved on the tangle

broker-node/app/DataMapping.php
AaronVasquez: It's generated on both the client and server.

@pandora23 what's the reasoning for this? This introduced complexity (they both have to be in sync) and race conditions (they assume that the other has the hashes already generated)

broker-node/app/DataMapping.php
AaronVasquez: Not sure why we do this tbh, the DB is still hit N times. 

@pandora23 Are we trying to batch hash generation? It looks like we just loop through them all too. Why do we need to batch them?

broker-node/app/Http/Controllers/UploadSessionController.php
AaronVasquez: It follows the ActiveRecord pattern. It has methods attached that can persist the models.

broker-node/app/DataMapping.php
EdmundMai: Isn't the hash generated on the client side? hash == encrypted file chunk, correct?

PR #19, opened 2018-01-15
Send chunks to hook node

AaronVasquez:


AaronVasquez:
@rfornea trying to understand the requirements. So the brokernode calls prepareTransfers() using an IOTA client in order to get the params the hook node needs?

Do we have a php IOTA client already?

PR #20, opened 2018-01-16
Chunk polling endpoint

AaronVasquez:
- Adds status field to data_maps table
- Adds endpoint to poll for a chunk's status
- Diff is a little noisy because DataMapping is renamed to DataMap

### DIFF
--- a/broker-node/app/Http/Controllers/UploadSessionController.php
+++ b/broker-node/app/Http/Controllers/UploadSessionController.php
@@ -122,8 +122,11 @@ class UploadSessionController extends Controller
      */
     public function chunkStatus(Request $request)
     {
+        $genesis_hash = $request->input('genesis_hash');
+        $chunk_idx = $request->input('chunk_idx');
+
         $data_map = DataMap::where('genesis_hash', $genesis_hash)
-            ->where('chunk_idx', $chunk['idx'])
+            ->where('chunk_idx', $chunk_idx)
             ->first();
         if (empty($data_map)) return response('Datamap not found', 404);
 
--- a/broker-node/app/Http/Controllers/UploadSessionController.php
+++ b/broker-node/app/Http/Controllers/UploadSessionController.php
@@ -2,7 +2,7 @@
 
 namespace App\Http\Controllers;
 
-use App\DataMapping;
+use App\DataMap;
 use App\UploadSession;
 use Illuminate\Http\Request;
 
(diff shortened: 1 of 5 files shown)


AaronVasquez:
Corresponding [webinterface PR](https://github.com/oysterprotocol/webinterface/pull/29)

broker-node/app/Http/Controllers/UploadSessionController.php
```diff
@@ -112,4 +112,24 @@ public function destroy($id)
...
+    {
+        $genesis_hash = $request->input('genesis_hash');
+        $chunk_idx = $request->input('chunk_idx');
+
+        $data_map = DataMap::where('genesis_hash', $genesis_hash)
+            ->where('chunk_idx', $chunk_idx)
+            ->first();
+        if (empty($data_map)) return response('Datamap not found', 404);
+
+        return response()->json(['status' => $data_map['status']]);
```
AaronVasquez: @rfornea How will the status get set? Are we going to be polling the tangle?

rfornea: @AaronVasquez   Yes, we'll be calling the broker's IRI API method "findTransactions" to see if anything is on the tangle for a particular address.  This won't tell us if it's confirmed or not but it will at least tell us if the attach was successful.

AaronVasquez: How long does the call take? Can we call this on the spot or do we need to poll for this in the background?

Also, if we just do this on the spot, can the client check directly or does the broker node also need to know about the status?

rfornea: That call is usually pretty short, less than 10 seconds.  I am open to suggestions here, I guess what we'd do is (for a particular chunk's intended address on the tangle) make the call repeatedly and wait for the result to return each time, and continue doing that or X number of times/Y number of seconds, or until we get back the result we want, whichever comes first.  

For the testnet I think we could "get away" with doing this on the spot but it would definitely be better if this was something that was happening in the background.  

Both the broker and the client have to know about the status.  The broker has to check because the hook node could be lying.  The broker will tell the client (when the client polls) the result of the attach, but the broker could be lying too, so the client has to do its own check--and when the client does its own check, it is supposed to be connecting to the tangle using a different node than the one that claims the attach was successful.  

We may want to simplify something of this for the testnet.  @nymd thoughts?

nymd: While the active polling does sound inefficient, let's do it the simplest way first. If we find the 10-second calls are the bottleneck once the system works, we can refactor to a background process.

AaronVasquez: Ok, I will do that in a separate PR so I can integrate with @rfornea's changes

PR #21, opened 2018-01-18
Rebel dev

rfornea:
These changes will need to be integrated with Aaron's changes.  

This review adds method that allows us to pass an address, message, and chunk id to the broker node.  Broker node will execute:
-findTransactions (see if tangle already has something at that address) (IRI method)
-prepareTransfers (build the tryte string from the tx data)
-getTransactionsToApprove (get the PoW the tangle wants us to do) (IRI method)
-sendToHookNode (sends everything to the hook node, which is currently the same node as the broker node but is listening on a different port)

There are also places stubbed out for other people's work to be inserted (or just replace the methods that are in there with different methods).


html/BrokerNode.php
```diff
@@ -0,0 +1,227 @@
...
+require_once("requests/IriData.php");
+require_once("requests/NodeMessenger.php");
+require_once("iota-php-port/PrepareTransfers.php");
+
+class BrokerNode
+{
+    public static $chunksToAttach = null;
+    public static $chunksToVerify = null;
+
+    public static $IriWrapper = null;
```
AaronVasquez: Can we set this to `new IriWrapper()` here? Lazy instantiation doesn't seem necessary.

html/BrokerNode.php
```diff
@@ -0,0 +1,227 @@
...
+    }
+
+    public static function getTransactionData(&$request)
+    {
+        $trytesToBroadcast = NULL;
+
+        $request->value = IriData::$txValue;
+        $request->tag = IriData::$oysterTag;
+
+        try {
```
AaronVasquez: This can be simplified (same way as your earlier PR)

html/BrokerNode.php
rfornea: My IDE gives me this error if I try to do that:  "expression is not allowed as field default value."  I think it's because it's static.  

We could change the variables and methods in this file to not be static if we wanted, but I was trying to not instantiate BrokerNode, IriWrapper, and NodeMessenger over and over, since we'll be calling them thousands of times.  The approach I've taken here might not be the best way to do it and we could revisit it.  

I am going to make the other change though.

html/BrokerNode.php
rfornea: I will change this.

html/BrokerNode.php
AaronVasquez: Ok, let's revisit later. I'm not a PHP expert, so I don't know what the best solution would be

PR #23, opened 2018-01-19
Integrate hooknode with laravel

AaronVasquez:
- This does **not** change behavior, just simply moves over logic into laravel app
- Some hacks were used in order to make integration easier. We can come back to this after testnet.
- Most of the changes were moving files around. Please review:
    - broker-node/app/Clients/BrokerNode.php
    - broker-node/routes/api.php
    - html/BrokerListener.php (deleted this and moved this logic)
    - broker-node/app/Http/Controllers/UploadSessionController.php


broker-node/app/Http/Controllers/UploadSessionController.php
```diff
@@ -140,7 +140,7 @@ public function brokerNodeListener(Request $request)
         // This is a hack to cast an associative array to an object.
         // I don't know how to use PHP properly :(
         $req = (object)[
-            "command" => $request->input('command'),
+            "command" => $cmd,
             "message" => $request->input('message'),
```
rfornea: $request->input('command') did not work?  Weird.

broker-node/app/Http/Controllers/UploadSessionController.php
```diff
@@ -132,4 +133,38 @@ public function chunkStatus(Request $request)
...
+    public function brokerNodeListener(Request $request)
+    {
+        $cmd = $request->input('command');
+        // This is a hack to cast an associative array to an object.
+        // I don't know how to use PHP properly :(
+        $req = (object)[
+            "command" => $request->input('command'),
+            "message" => $request->input('message'),
+            "chunkId" => $request->input('chunkId'),
+            "address" => "WHLOOOOOOAAAAAAAAAAAAAAAAALAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA",
```
rfornea: Do you know if we already have the functionality to generate the address from the chunk's entry in the data map?  We will have to remember to modify this function to get the address that way, once we have the ability to do so.  

@pandora23  Do we already have methods yet that will generate the 81 character tryte address from a chunk's hash in the data map?

PR #25, opened 2018-01-19
Process chunk endpoint

AaronVasquez:


broker-node/app/Clients/BrokerNode.php
```diff
@@ -41,6 +41,7 @@ private static function initMessenger()
 
     public static function processNewChunk(&$chunk)
     {
+        $chunk->command = "processNewChunk";
```
rfornea: We may not need this anymore.  If the API is now changed to not check for particular command strings we can probably get rid of this line.  But we don't necessarily need to, it's not hurting anything.

AaronVasquez: If it's not needed, its less confusing to just delete it.

PR #26, opened 2018-01-19
Check tangle for chunk status

AaronVasquez:


broker-node/app/Http/Controllers/UploadSessionController.php
```diff
@@ -131,6 +134,28 @@ public function chunkStatus(Request $request)
             ->first();
         if (empty($data_map)) return response('Datamap not found', 404);
 
+        // Error Responses
+        if (empty($data_map)) {
```
rfornea: Lines 135 and 138 seem to be the same?

PR #32, opened 2018-01-24
Confirm chunk

rfornea:
Disregard d221d65

This gives us the ability to verify if a chunk on the tangle has the same message, trunk transaction, and branch transaction that we have in our records.


rfornea:
Okay.  I'll go ahead and push it in and try to make that section clearer in my next push.

PR #37, opened 2018-01-25
No more hardcoding.

AaronVasquez:


broker-node/app/Http/Controllers/UploadSessionController.php
```diff
@@ -40,8 +40,8 @@ public function store(Request $request)
         $genesis_hash = $request->input('genesis_hash');
         $file_size_bytes = $request->input('file_size_bytes');
 
-        // TODO: Where does $file_chunk_count this come from.
-        $file_chunk_count = 3;
+        // TODO: Make this an env variable.
+        $file_chunk_count = $file_size_bytes / 1000;
```
EdmundMai: we need to round this up:

print (4500/1000); => 4.5 // we want 5
Playground: https://www.tehplayground.com/a1hlo1q2UJqsBzts

PR #38, opened 2018-01-26
Client broker integration v2

EdmundMai:


broker-node/app/Clients/requests/IriData.php
```diff
@@ -4,7 +4,7 @@
 
 class IriData
 {
-    public static $nodeUrl = "http://localhost:14265";
+    public static $nodeUrl = "http://165.227.79.113:14265";
```
rfornea: This probably should be localhost.  Usually, the iri requests are going to be coming from within the broker.

broker-node/app/Http/Controllers/UploadSessionController.php
```diff
@@ -97,20 +98,22 @@ public function update(Request $request, $id)
...
-        if (empty($data_map)) {
-            return response('Datamap not found', 404);
-        }
-        if ($data_map['hash'] != $chunk['hash']) {
-            return response('Forbidden', 403);
-        }
+        if (empty($data_map)) return response('Datamap not found', 404);
+
+        $trytes = new Trytes(["characters" => Trytes::IOTA]);
+
```
rfornea: When I tested Tuupola/trytes, I tested it without passing a param to the constructor.  It's probably fine this way, but just wanted to note that in case we see weird results later.

broker-node/app/Clients/BrokerNode.php
```diff
@@ -41,21 +43,15 @@ private static function initMessenger()
...
-                // move on to the next chunk
-                /*
-                 * TODO: is there anything specific we want to do if it's
-                 * already attached?
-                 */
-            }
-        } catch
-        (\Exception $e) {
-            echo "Caught exception: " . $e->getMessage();
-            // something went wrong during our check, do something about it
```
EdmundMai: let's stop doing this "try / catch error and do nothing" pattern in the future

broker-node/app/Clients/BrokerNode.php
```diff
@@ -132,7 +124,7 @@ private static function selectHookNode()
         For now, we have limited nodes so we are just hard-coding nodes
 
         */
-        return "http://localhost:250";
+        return "http://165.227.79.113:250";
```
EdmundMai: should this stay as localhost?

broker-node/app/Clients/iota-php-port/kerl-php/conv.php
```diff
@@ -58,8 +26,35 @@ function trytes_to_trits($trytes)
 
 function trits_to_trytes($trits)
 {
-    global $trit_table;
-    global $tryte_table;
+    $tryte_table = array(
```
EdmundMai: was NULL using the global var strategy above. we can figure out why post-testnet A

broker-node/app/Clients/iota-php-port/kerl-php/conv.php
```diff
@@ -58,8 +26,35 @@ function trytes_to_trits($trytes)
 
 function trits_to_trytes($trits)
 {
-    global $trit_table;
```
EdmundMai: not being used

broker-node/app/Clients/BrokerNode.php
pandora23: Logging is usually appropriate.  Not sure if we ever decided on a logger.

PR #44, opened 2018-01-27
Looks for wrong status.

AaronVasquez:


pandora23:
Could you comment the purge method?  Not sure what it is doing.  Selecting rows with status complete and then ?

PR #45, opened 2018-01-27
Adds example env file

AaronVasquez:


broker-node/env-example
```diff
@@ -0,0 +1,35 @@
+APP_NAME=broker-node
+APP_ENV=local
+APP_KEY=base64:jYKeH2xka3+sB5mn1e27bFGU/49vr1qK7JJuHO5hMZs=
```
nymd: This should probably default as blank

PR #51, opened 2018-01-28
Register hooknode

AaronVasquez:


broker-node/app/Http/Controllers/HookNodeController.php
```diff
@@ -0,0 +1,23 @@
...
+{
+
+    /**
+     * Store a newly created resource in storage.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @return \Illuminate\Http\Response
+     */
+    public function store(Request $request) {
+        $ip_address = $request->input('ip_address');
```
AaronVasquez: @rfornea @pandora23  We could use $request->ip() instead of passing it as a param, but that wouldn't work if the hook node sat behind a proxy. Would there be any incentive for a hoodnode to send an invalid ip_address?

rfornea: A hooknode wouldn't want to send an invalid IP, but other brokernodes might--in order to screw up other brokernodes and ruin their reputation by sending them hooknodes that aren't any good or aren't real.

AaronVasquez: Good point, other brokernodes can "DDOS" it with bad IPs. I don't think the reputation system can protect against that. This might be something to think about after testnet

@brunoblock how can we protect against this?

PR #54, opened 2018-01-29
Feature/cron job

rfornea:
I cannot verify yet that these changes work the way I expect because of timeout issues but this set of changes also includes some other changes we are going to want anyway.  Caught them while working on cron stuff.


broker-node/app/Clients/BrokerNode.php
```diff
@@ -172,7 +169,16 @@ private static function updateHookNodeDirectory($currentHook, $status)
...
 
-    public static function verifyChunkMatchesRecord($chunk)
+    public static function verifyChunkMessageMatchesRecord($chunk)
+    {
+        try {
+            return self::verifyChunkMatchesRecord($chunk, false);
+        } catch (\Exception $e) {
+            echo "Caught exception: " . $e->getMessage() . $GLOBALS['nl'];
+        }
+    }
```
rfornea: I think Edmund didn't like the try/catch pattern, should I change this?

AaronVasquez: Yeah, if an exception is thrown, we should let it bubble up unless we handle it. Silent errors might cause bigger problems since we won't be aware of it.

We now have Sentry to catch them, and I added middleware to respond to the client with the exception if its uncaught.

rfornea: Okay I'll take it out

PR #57, opened 2018-01-29
Prepare transfers fixes

rfornea:
This was throwing an error

### DIFF
--- a/broker-node/app/Clients/iota-php-port/PrepareTransfers.php
+++ b/broker-node/app/Clients/iota-php-port/PrepareTransfers.php
@@ -24,8 +24,15 @@ class PrepareTransfers
         // Also remove the checksum of the address if it's there after validating it
         foreach ($transfers as $thisTransfer) {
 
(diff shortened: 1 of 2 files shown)


broker-node/app/Clients/iota-php-port/PrepareTransfers.php
```diff
@@ -24,8 +24,15 @@ public static function prepareTransfers($seed, $transfers, $options = null, $cal
         // Also remove the checksum of the address if it's there after validating it
         foreach ($transfers as $thisTransfer) {
 
-            $thisTransfer->message = $thisTransfer->message ? $thisTransfer->message : '';
-            $thisTransfer->obsoleteTag = $thisTransfer->tag ? $thisTransfer->tag : ($thisTransfer->obsoleteTag ? $thisTransfer->obsoleteTag : '');
```
nymd: Thanks for expanding this.
I don't mind the short if/then notation if it is just one but nested it gets hard to understand at a glance.

PR #59, opened 2018-01-29
Get next hooknode

AaronVasquez:


broker-node/app/Http/Controllers/UploadSessionController.php
```diff
@@ -117,22 +117,26 @@ public function update(Request $request, $id)
...
-            $data_map->message = $message_in_tryte_format;
-            $data_map->chunk = $chunk["data"];
-            $data_map->status = DataMap::status['pending'];
-            $data_map->save();
-
-            return response('Success.', 204);
+        [$res_type, $updatedChunk] = BrokerNode::processNewChunk($brokerReq);
+
+        switch($res_type) {
+            case 'already_attached':
```
AaronVasquez: @rfornea this changes what you did. Can you please make sure I understood what you were doing and changed it correctly?

rfornea: This looks like it will be fine.  As long as it's still returning all the data that's on a chunk, and it looks like it is, we should be good to go.

PR #75, opened 2018-01-31
Remove tryte conversion on brokernode (relies on client)

AaronVasquez:


broker-node/app/Http/Controllers/UploadSessionController.php
```diff
@@ -102,16 +101,9 @@ public function update(Request $request, $id)
...
-
-        $message_in_tryte_format = $trytes->encode($chunk["data"]);
-
-        $hash_in_tryte_format = $trytes->encode($data_map["hash"]);
-        $shortened_hash = substr($hash_in_tryte_format, 0, 81);
-
         // Save address and message on data_map
-        $data_map->address = $shortened_hash;
-        $data_map->message = $message_in_tryte_format;
+        $data_map->address = $data_map["hash"];
```
AaronVasquez: @rfornea the `address` field might no longer be needed, but let's make sure this works before doing that bigger change.

PR #92, opened 2018-02-06
Simple hook queue

rfornea:


broker-node/app/Clients/BrokerNode.php
```diff
@@ -45,249 +45,21 @@ private static function initMessenger()
         }
     }
 
-    private static function getOrInitHooknodeQueue()
+    private static function getNextXHookNodes($numberOfHooks)
     {
-        if (!is_null(self::$hooknode_queue)) {
-            return self::$hooknode_queue;
-        }
+        $arrayQueue = file(__DIR__ . "/" . "HookNodes.txt", FILE_IGNORE_NEW_LINES);
```
AaronVasquez: This is a race condition where concurrent threads are all reading and writing to the same file. 

I think we'll run into the same issue as before -- when the client sends chunks super fast, the same first X hooknodes will be used since the threads don't wait for the previous one to finish writing before they read from the file.

PR #95, opened 2018-02-07
Integrate multi chunk

AaronVasquez:


broker-node/app/Clients/BrokerNode.php
```diff
@@ -380,25 +375,24 @@ public static function buildTransactionData($chunks)
...
         $result = self::$IriWrapper->makeRequest($command);
 
-        BrokerNode::$iriRequestInProgress = false;
 
         if (!is_null($result) && property_exists($result, 'branchTransaction')) {
             //switching trunk and branch
             //do we do this randomly or every time?
+            $tmp_trunk = $request->trunkTransaction;
             $request->trunkTransaction = $result->branchTransaction;
-            $request->branchTransaction = $result->trunkTransaction;
```
AaronVasquez: @rfornea I think this was confirming the branch as both branch and trunk? Not sure if these might have caused any slowdowns or problems?

rfornea: It just swaps them.  Something about entropy.

broker-node/app/Clients/BrokerNode.php
```diff
@@ -291,18 +290,16 @@ public static function processChunks(&$chunks)
         }
     }
 
-    public static function filterUnattachedChunks($addresses, $chunks)
+    private static function filterUnattachedChunks($addresses, $chunks)
```
rfornea: Thanks for changing all the other methods to private.  This one we might need to keep public though?  I think CheckChunkStatus will need to use either this, or the verifyChunkMatchesRecord.  Speaking of which, I need to change that method to accept an array as well.   

We can leave it like this for now though, there's no cron task currently afaik.

broker-node/app/Clients/BrokerNode.php
```diff
@@ -380,25 +375,24 @@ public static function buildTransactionData($chunks)
...
 
-        BrokerNode::$iriRequestInProgress = false;
 
         if (!is_null($result) && property_exists($result, 'branchTransaction')) {
             //switching trunk and branch
             //do we do this randomly or every time?
+            $tmp_trunk = $request->trunkTransaction;
             $request->trunkTransaction = $result->branchTransaction;
-            $request->branchTransaction = $result->trunkTransaction;
+            $request->branchTransaction = $tmp_trunk;
```
rfornea: Unless this is some assigning by reference, I'm not sure this will work?  $request->trunkTransaction is undefined at the time tmp_trunk is assigned.  $request->trunkTransaction and $request->branchTransaction are both empty, and get assigned to the values of $result->trunkTransaction and $result->branchTransaction

AaronVasquez: Reverted, I misread this as as swapping instead of assigning to a $request object.

broker-node/app/Clients/BrokerNode.php
AaronVasquez: Reverted.

PR #97, opened 2018-02-08
going back to eclipse to push

pandora23:
Ok.  Getting tired of trying to get the VM git solution working.  
Pushing from eclipse again.

### DIFF
--- a/broker-node/app/ChunkEvents.php
+++ b/broker-node/app/ChunkEvents.php
@@ -1,0 +1,89 @@
+<?php
+
+
+namespace App;
+
+
+
+use App\Clients\BrokerNode;
+
+use Illuminate\Database\Eloquent\Model;
+
+use Webpatser\Uuid\Uuid;
+
+
+
+class ChunkEvents extends Model
+
+{
+
+
+    public static function boot()
+    {
+
+        parent::boot();
+
+        self::creating(function ($model) {
+
+            $model->id = (string)Uuid::generate(4);
+
+        });
+
+    }
+    
+    protected $table = 'chunk_events';
+    protected $fillable = [
+        'hook_node_id',
+        'session_id',
+        'event_name',
+        'value'
+    ];
+    public $incrementing = false;  // UUID
+    
+
+	
+	public static function addChunkEvent($event_name, $hook_node_id, $session_id, $value){
+	
+		self::create([
+
+		    'hook_node_id' => $hook_node_id,
+		    
+		    'session_id' => $session_id,
+
+		    'event_name' => $event_name,
+
+		    'value' => $value
+
+		]);
+		
+		//$this->save();
+
+	}
+	
+	//refactor   setting the hook id to be the ip.  
+	//TODO:  add session id
+	public function addChunkSentToHookNodeEvent($hook_ip){
+	    
+	    $event_name = "chunk_sent_to_hook";
+	     
+	    self::addChunkEvent($event_name, $hook_ip, "", "" );
+	}
+	
+	public function addHookNodeFinishedChunkEvent($hook_ip){
+	    
+	    $event_name = "chunk_finished_by_hook";
+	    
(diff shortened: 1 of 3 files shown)


broker-node/app/ChunkEvents.php
```diff
@@ -0,0 +1,89 @@
...
+    ];
+    public $incrementing = false;  // UUID
+    
+
+	
+	public static function addChunkEvent($event_name, $hook_node_id, $session_id, $value){
+	
+		self::create([
+
+		    'hook_node_id' => $hook_node_id,
```
rfornea: This is super nit-picky, but it might be better to call this 'hooknode_id' instead of 'hook_node_id'.  hooknode_id is what we are using elsewhere (in the DataMap and CheckChunkStatus files).  Alternatively, we should change it to hook_node_id in those other places.

broker-node/app/ChunkEvents.php
```diff
@@ -0,0 +1,89 @@
+<?php
+
+
+namespace App;
+
```
AaronVasquez: Nit: Don't double space code. Although we don't have a style guide yet for PHP, code should be somewhat consistent.

broker-node/app/ChunkEvents.php
```diff
@@ -0,0 +1,89 @@
...
+        'hook_node_id',
+        'session_id',
+        'event_name',
+        'value'
+    ];
+    public $incrementing = false;  // UUID
+    
+
+	
+	public static function addChunkEvent($event_name, $hook_node_id, $session_id, $value){
```
AaronVasquez: Make this `private`. 

Also, not sure if this abstraction is needed. It becomes less clear which values are saved since you rely on parameter position rather than an associative array.

broker-node/app/ChunkEvents.php
```diff
@@ -0,0 +1,89 @@
...
+
+		]);
+		
+		//$this->save();
+
+	}
+	
+	//refactor   setting the hook id to be the ip.  
+	//TODO:  add session id
+	public function addChunkSentToHookNodeEvent($hook_ip){
```
AaronVasquez: These should be `static` since you aren't using any instance variables.

broker-node/app/Clients/BrokerNode.php
```diff
@@ -37,6 +40,13 @@ private static function initIri()
         }
     }
 
+	private static function initEventRecord()
```
AaronVasquez: You can remove this if you make those functions above `static`

broker-node/app/Clients/BrokerNode.php
```diff
@@ -428,6 +438,10 @@ private static function sendToHookNode(&$chunks, $request)
         self::$NodeMessenger->spamHookNodes($tx, $spammedNodes);  // remove this, temporary solution
 
         self::updateHookNodeDirectory($hookNodeUrl, "request_made");
+		
+		//store the event record for analysis
+		self::initEventRecord();
+        self::$ChunkEventsRecord->addChunkSentToHookNodeEvent($hooknode['ip_address']);
```
AaronVasquez: This becomes `ChunkEvent::addChunkSentToHookNodeEvent($hooknode['ip_address']);` if you make that method static.

broker-node/app/ChunkEvents.php
```diff
@@ -0,0 +1,89 @@
...
+
+use App\Clients\BrokerNode;
+
+use Illuminate\Database\Eloquent\Model;
+
+use Webpatser\Uuid\Uuid;
+
+
+
+class ChunkEvents extends Model
```
AaronVasquez: Nit: Models are typically singular (ChunkEvent) while tables and controllers are plural (chunk_events, ChunkEventsController)

pandora23: k
I will look through everything and conform to standards.  Wouldn't mind a standards guide.

pandora23:
Using the next system

PR #110, opened 2018-02-10
Update 2018_01_28_155850_create_hook_nodes_table.php

rfornea:


broker-node/database/migrations/2018_01_28_155850_create_hook_nodes_table.php
```diff
@@ -31,7 +31,6 @@ public function up()
 
             // Indexes
             $table->primary('id');
-            $table->unique(['ip_address']);
```
AaronVasquez: When would this not be unique?

AaronVasquez: @rfornea @pandora23 ^

PR #125, opened 2018-02-20
Rank hooknodes

AaronVasquez:


rfornea:
This looks good.  I guess we can add the BrokerNode querying the HookNode's load later.  I would lean towards wanting to make those calls from within getNextReadyNode(), before we ever return the hooknode.

PR #126, opened 2018-02-21
Tip selection (don't deploy to brokers until new IRI is installed)

rfornea:
-adds GetFreshTips command and adds it to cron job
-only stores unique tips, or updates existing tips if they still get returned from the getBulkTransactionsToApprove method
-makes sure the pair of tips comes from the same node otherwise we'll be pulling from opposite sides of the tangle
-increase depth to 4 to get more unique transactions
-uses DB for trunk and branch unless the DB is empty


broker-node/app/Clients/BrokerNode.php
```diff
@@ -183,21 +165,26 @@ private static function buildTransactionData($chunks)
 
     private static function getTransactionsToApprove(&$request)
     {
-        self::initIri();
+        $tips = Tips::getNextTips();
 
-        $command = new \stdClass();
-        $command->command = "getTransactionsToApprove";
-        $command->depth = IriData::$depthToSearchForTxs;
+        if ($tips != null) {
```
AaronVasquez: You should also check if `$tips[0]` and `$tips[1]` are not null.

broker-node/app/Clients/BrokerNode.php
```diff
@@ -10,45 +10,30 @@
...
 class BrokerNode
 {
     public static $chunksToAttach = null;
     public static $chunksToVerify = null;
 
-    public static $IriWrapper = null;
-    public static $NodeMessenger = null;
     public static $ChunkEventsRecord = null;
 
-    private static function initIri()
```
AaronVasquez: Nice! Making these static is much cleaner 👍

PR #130, opened 2018-02-26
Testnet a patches

rfornea:
-adds a getBroadcastingNodes method to HookNode.  Selects hooknodes similar to getNextReadyNode but with a lower score weight.  Does not update 'contacted_at' because we don't want the hooknodes to miss their chance of increasing their score just because they did a broadcast 
-passes that array of hooknodes to the hooknode who will perform attachToTangle.
-disables the TipSelection service.  It seems to be doing more harm than good.  It's slowing the broker down a lot, and the new changes to iota seem to be resulting in better tips.  
-in the cron job, only re-send chunks in batches of 10.

### DIFF
--- a/broker-node/app/Console/Kernel.php
+++ b/broker-node/app/Console/Kernel.php
@@ -28,9 +28,9 @@ class Kernel extends ConsoleKernel
             ->everyMinute()
             ->sendOutputTo('/var/CHUNKLOG');
 
-        $schedule->command('GetFreshTips:getTips')
-            ->everyFiveMinutes()
-            ->sendOutputTo('/var/TIPSLOG');
+//        $schedule->command('GetFreshTips:getTips')
+//            ->everyFiveMinutes()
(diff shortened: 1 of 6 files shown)


broker-node/app/Console/Kernel.php
```diff
@@ -28,9 +28,9 @@ protected function schedule(Schedule $schedule)
             ->everyMinute()
             ->sendOutputTo('/var/CHUNKLOG');
 
-        $schedule->command('GetFreshTips:getTips')
-            ->everyFiveMinutes()
-            ->sendOutputTo('/var/TIPSLOG');
+//        $schedule->command('GetFreshTips:getTips')
```
AaronVasquez: Did you mean to keep this commented out?

rfornea: Yeah.  I will double check later tonight, but I think the tip selection service is doing more harm than good.

PR #132, opened 2018-02-27
Installs segment client with sample log

AaronVasquez:


broker-node/app/Clients/BrokerNode.php
```diff
@@ -212,8 +212,17 @@ private static function sendToHookNode(&$chunks, $request)
         NodeMessenger::sendMessageToNodesAndContinue($tx, $hookNodes);
 
         //record event
-        self::initEventRecord();
         self::$ChunkEventsRecord->addChunkEvent("chunk_sent_to_hook", $hookNodeUrl, "todo", "todo");
+        Segment::track([
+            "event" => "chunk_sent_to_hook",
+            "properties" => [
+                "broker_url" => $_SERVER['REMOTE_ADDR'],
```
rfornea: For this, if we want to add the chunk id as one of the properties, we could move these lines of code a few lines down, into the array_walk statement, where we would have access to each chunk's id.

AaronVasquez: I'll leave this for a separate PR. I'm about to leave for lunch and don't want to block you from creating more logs.

PR #133, opened 2018-02-27
Adds Full Segment Logs

rfornea:
Adds these to logs to the brokernode code, with Segment:  

1.  Chunk confirmed with the hook node id and chunk id  (cron job)
2.  Chunk sent to hook node (brokernode.php)
3.  Chunk sent from client.  (can't remember, its in there)
4.  Any 'decrement' of rep score.  (will help us identify  bad hooks)
5.  Chunk resent maybe.  Any failures.


broker-node/app/Clients/BrokerNode.php
```diff
@@ -27,10 +28,13 @@ class BrokerNode
 
     public static $ChunkEventsRecord = null;
 
-    private static function initEventRecord()
+    private static $SegmentStarted = null;
```
AaronVasquez: We can add `AltThree\Segment\SegmentServiceProvider::class,` on line `171` in `broker-node/config/app.php`

Sorry, forgot to initialize it.

broker-node/app/Clients/BrokerNode.php
```diff
@@ -27,10 +28,13 @@ class BrokerNode
...
+
+    private static function initSegment()
     {
-        if (is_null(self::$ChunkEventsRecord)) {
-            self::$ChunkEventsRecord = new ChunkEvents();
+        if (is_null(self::$SegmentStarted )) {
+            Segment::init("SrQ0wxvc7jp2XDjZiEJTrkLAo4FC2XdD");
```
AaronVasquez: Shouldn't be committing this key into Github. If you add it to `broker-node/.env` as `SEGMENT_WRITE_KEY`

broker-node/config/app.php
```diff
@@ -177,6 +177,7 @@
         // App\Providers\BroadcastServiceProvider::class,
         App\Providers\EventServiceProvider::class,
         App\Providers\RouteServiceProvider::class,
+        AltThree\Segment\SegmentServiceProvider::class
```
AaronVasquez: Nit: move this to line `171` since its' a package service provider.

Do you still need to call `Segment::init("SrQ0wxvc7jp2XDjZiEJTrkLAo4FC2XdD");` with this and the `SEGMENT_WRITE_KEY` ENV variable added? I think the framework should have taken care of it?

broker-node/config/segment.php
```diff
@@ -24,6 +24,5 @@
     |
     */
 
-    'write_key' => env('SEGMENT_WRITE_KEY'),
-
+    'write_key' => env('SrQ0wxvc7jp2XDjZiEJTrkLAo4FC2XdD'),
```
AaronVasquez: Shouldn't be committing this key into Github. This will read `SEGMENT_WRITE_KEY` from `broker-node/.env`

broker-node/app/Clients/BrokerNode.php
AaronVasquez: Nvm, I see you already did it. The `ENV` variable just wasn't set up, which is why you might have been seeing errors.

broker-node/app/Http/Controllers/UploadSessionController.php
```diff
@@ -118,6 +124,16 @@ public function update(Request $request, $id)
                     ->where('chunk_idx', $chunk['idx'])
                     ->select('hash')
                     ->first();
+
+                Segment::track([
+                    "userId" => "Oyster",
```
AaronVasquez: Maybe we should use the server's IP as the `userId` ?

broker-node/app/Clients/BrokerNode.php
rfornea: I added it there, but could not get it to work.

broker-node/config/app.php
rfornea: I thought so too.

broker-node/app/Http/Controllers/UploadSessionController.php
rfornea: I would be okay with that.  $_SERVER['REMOTE_ADDR'] doesn't give us the broker's IP though, it gives us the client's.  I will see if there is some global variable we can use that would have the broker's IP.

rfornea: I think 'SERVER_ADDR' will work.

rfornea: SERVER_ADDR did not work.

broker-node/app/Clients/BrokerNode.php
```diff
@@ -227,21 +221,18 @@ private static function sendToHookNode(&$chunks, $request)
...
-        ]);
-
-        // DEPRECATED. This will be replaced with segment.io
-        self::initEventRecord();
-        self::$ChunkEventsRecord->addChunkEvent("chunk_sent_to_hook", $hookNodeUrl, "todo", "todo");
         HookNode::incrementChunksProcessed($hookNodeUrl, count($chunks));
 
         array_walk($chunks, function ($chunk) use ($hookNodeUrl, $request) {
+            //record event
+            Segment::track([
```
AaronVasquez: Since we send this as an array, should we log just 1 event with an array of `chunk_idxs`? It would more accurately reflect what we send to the hooknode.

PR #153, opened 2018-03-14
Data maps

AaronVasquez:


models/data_maps.go
```diff
@@ -0,0 +1,93 @@
...
+func (d DataMap) String() string {
+	jd, _ := json.Marshal(d)
+	return string(jd)
+}
+
+// DataMaps is not required by pop and may be deleted
+type DataMaps []DataMap
+
+// String is not required by pop and may be deleted
+func (d DataMaps) String() string {
```
rfornea: Do we need both this function as well as the one that is on line 33?

AaronVasquez: Notice how the one on `33` is singular and this one is plural. Since golang refuses to support generics, we have hacks like this :(

rfornea: Ah, I see.

PR #159, opened 2018-03-19
Golang cron tasks

rfornea:
This took quite a while because I was still getting the hang of the syntax, how to use the test suites, how to use the ORM, etc.  Subsequent cron tasks will go much faster.   

This is the method that will check the DB for unassigned chunks and process them.  Not sure whether we'll need this method now if we are getting rid of the hooknodes but this was the first one I started before I learned we were going to switch up the architecture.


models/data_maps.go
```diff
@@ -29,6 +31,20 @@ type DataMap struct {
 	Address     string    `json:"address" db:"address"`
 }
 
+func init() {
+	SetChunkStatuses()
+}
+
+func SetChunkStatuses() {
```
AaronVasquez: We can use native golang enums like this: https://github.com/oysterprotocol/brokernode/blob/da5d641488ecda2552fe03ed917c701db31a9487/models/upload_sessions.go#L13

So we have compiler type checks instead of strings

AaronVasquez: I would also start from int 1 instead of 0 since golang treats 0 as an empty value, we might get unexpected behavior

jobs/process_unassigned_chunks.go
```diff
@@ -0,0 +1,74 @@
...
+	if err != nil {
+		raven.CaptureError(err, nil)
+	} else {
+		processChunks(chunks)
+	}
+}
+
+func GetUnassignedChunks() (dataMaps []models.DataMap, err error) {
+
+	tx, err := pop.Connect("test")
```
AaronVasquez: This won't work in a non test environment. You might be able to access the `DB` ? (not sure where it was injected for the stuff I built). Otherwise, you will have to pass in the `ctx` object when you initialize the job.

models/data_maps.go
rfornea: Cool, I'll change this.

jobs/process_unassigned_chunks.go
rfornea: I meant to pull this out but forgot.

models/data_maps.go
```diff
@@ -6,16 +6,22 @@ import (
...
-
 const fileBytesChunkSize = float64(2817)
 
+const (
+	Pending int = iota + 1
+	Unassigned
+	Unverified
+	Complete
+	Confirmed
+	Error
```
AaronVasquez: Is it possible to make `Error` -1?

rfornea: Sure.  So far we haven't used error, and I'm not sure we even need it.  But I will change it.

PR #163, opened 2018-03-20
Adds "purgeCompletedSessions" job

rfornea:
-Adds "PurgeCompletedSessions" job and tests
-Moves completed data maps to a completed_data_maps table.  Will need this if we decide we care about confirmation rates, though we might not care anymore
-Deletes completed sessions

### DIFF
--- a/jobs/purge_completed_sessions.go
+++ b/jobs/purge_completed_sessions.go
@@ -4,6 +4,7 @@ import (
 	"github.com/oysterprotocol/brokernode/models"
 	"strconv"
 	"log"
+	"github.com/gobuffalo/pop"
 )
 
 func init() {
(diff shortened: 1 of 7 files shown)


jobs/purge_completed_sessions.go
```diff
@@ -1 +1,75 @@
...
+	for _, genesisHash := range genesisHashesNotComplete {
+		notComplete[genesisHash.GenesisHash] = true
+	}
+
+	var moveToComplete = []models.DataMap{}
+
+	for _, genesisHash := range allGenesisHashes {
+		if !notComplete[genesisHash] {
+			models.DB.RawQuery("SELECT * from data_maps WHERE genesis_hash = ?", genesisHash).All(&moveToComplete)
+			MoveToComplete(moveToComplete)
```
AaronVasquez: Wrap lines 48-50 in a transaction so we don't have inconsistent data.

models/completed_data_maps.go
```diff
@@ -0,0 +1,64 @@
...
+import (
+	"encoding/json"
+	"time"
+	"github.com/gobuffalo/pop"
+	"github.com/gobuffalo/validate"
+	"github.com/gobuffalo/validate/validators"
+)
+
+type CompletedDataMap struct {
+	ID          int       `json:"id" db:"id"`
```
AaronVasquez: You can merge fields from another struct like this:
```go
type CompletedDataMap struct {
    models.DataMap
}
```

rfornea: This doesn't seem to work, or at least I can't seem to make it work.  You can inherit from other structs, but the fields won't keep the same indentation level.  This demonstrates the issue:  https://play.golang.org/p/8VH15cNga9R  

Unless there is some other way to do it.

jobs/purge_completed_sessions.go
rfornea: Okay

models/completed_data_maps.go
AaronVasquez: ok, nvm then

PR #165, opened 2018-03-21
Adds a database for webnodes and a job "FlushOldWebnodes"

rfornea:
-adds "FlushOldWebnodes"
-adds a database for webnodes
-had to do some weird sleep thing in the tests.  You're supposed to be able to manually set updated_at with Exec statements.  I tried several things and was unsuccessful.

### DIFF
--- a/jobs/flush_old_webnodes.go
+++ b/jobs/flush_old_webnodes.go
@@ -1,0 +1,28 @@
+package jobs
+
+import (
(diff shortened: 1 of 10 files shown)


AaronVasquez:
The sleep call in the test isn't the best...but it's just a test, so yolo. We can clean it up if we learn a better way.

PR #170, opened 2018-03-22
Docker mysql

AaronVasquez:
Part 1/2 Sets up our current up to be dockerized, then we will add a private tangle instance in part 2/2.


rfornea:
I noticed the build ran a long time for this PR.  Is it going to be that long everytime if we add docker-compose?

AaronVasquez:
It got stuck in an infinite loop somewhere. Locally this took almost the same as the build before.

rfornea:
Okay.  Everything looks good to me.  I can approve once the infinite loop is fixed.

PR #174, opened 2018-03-23
Update README.md

AaronVasquez:


automyr:
We'll, they certainly don't make the review button easy to see on mobile browsers.

PR #179, opened 2018-03-28
Upload chunk

AaronVasquez:


actions/upload_sessions.go
```diff
@@ -82,6 +95,47 @@ func (usr *UploadSessionResource) Create(c buffalo.Context) error {
...
+
+			// Updates dmap in DB.
+			dm.Message = chunk.data
+			dm.Status = models.Unassigned
+			tx.ValidateAndSave(&dm)
+
+			dMaps[i] = dm
+		}
+
+		// Should we still do this here?
```
AaronVasquez: @rfornea Do we still want to process chunks here, or just let the cron job take care of this?

actions/upload_sessions.go
```diff
@@ -82,6 +95,47 @@ func (usr *UploadSessionResource) Create(c buffalo.Context) error {
...
+		dMaps := make([]models.DataMap, len(req.chunks))
+		for i, chunk := range req.chunks {
+			// Fetch DataMap
+			dm := models.DataMap{}
+			tx.RawQuery(
+				"SELECT * from data_maps WHERE genesis_hash ? AND chunk_idx = ?", uploadSession.GenesisHash, chunk.idx).First(&dm)
+
+			// Updates dmap in DB.
+			dm.Message = chunk.data
+			dm.Status = models.Unassigned
```
AaronVasquez: @rfornea Is `unassigned` the correct status here?

rfornea: I believe so.  As long as it has the message we should be good to go, address and message is what we need.

actions/upload_sessions.go
rfornea: We can pick it up in the cron job.  We can change the frequency of the cron task to run as often as we like.

rfornea:
I'm just gonna go ahead and approve and I'll pull those two lines out in my next PR.

PR #190, opened 2018-04-03
Changes that came up while QA'ing uploads to Golang broker.

rfornea:
-Some json mappings to structs didn't match what client was sending.  They match now.
-In the "Update" function of upload-sessions, getting the chunks from req.chunks didn't work until the structs involved were capitalized.  Not sure why.
-The transaction wrapping was giving me errors, something like "This transaction has already been rolled back or committed."  So for now just switch to models.DB
-fixed a URL
-added an ID to uploadSessionCreateRes because otherwise the ID was undefined for the client.

As far as I know, all of this should work.  I am getting as far as the broadcast step but calling IRI from within the docker container is giving me errors, despite using the IP we are supposed to use within containers.  I will look into this tomorrow.


models/chunk_channels.go
```diff
@@ -79,6 +79,21 @@ func GetReadyChannels() ([]ChunkChannel, error) {
 	return channels, err
 }
 
+// GetReadyChannels grabs one ready channel
+func GetOneReadyChannel() (ChunkChannel, error) {
+
+	channel := ChunkChannel{}
+
+	err := DB.RawQuery("SELECT * from chunk_channels WHERE "+
+		"est_ready_time <= ? ORDER BY est_ready_time;", time.Now()).First(&channel)
```
AaronVasquez: When is `est_ready_time` updated? Might want to atomically update it here (rename the method to `GetAndReserveChannel`) because concurrent threads might get the same channel from the DB.

Doesn't have to be this PR, but we should make an issue to get back to this if it becomes a problem.

rfornea: It gets updated in a method called SetEstimatedReadyTime in iota_wrappers.go  

Eventually I want to relocate it to the chunk_channels.go file but there was some kind of import cycle when I tried that, so for now just keeping it in the iota_wrappers file.

PR #191, opened 2018-04-04
Register webnode

EdmundMai:


migrations/schema.sql
```diff
@@ -0,0 +1,165 @@
+-- MySQL dump 10.16  Distrib 10.1.26-MariaDB, for debian-linux-gnu (x86_64)
```
rfornea: Do we need to check this file in?  @AaronVasquez do you know?

PR #194, opened 2018-04-05
Adds obfuscation to hashes

AaronVasquez:
Webinterface change coming soon.


models/data_maps.go
```diff
@@ -17,12 +20,12 @@ import (
 const fileBytesChunkSize = float64(2187)
 
 const (
-	Pending    int = iota + 1
+	Pending int = iota + 1
```
AaronVasquez: Our `gofmt` seem to keep disagreeing with each other. Which version are you using? I'm on 1.10

```bash
➜ go version
go version go1.10 darwin/amd64
```

rfornea: To be totally honest, a lot of times I just use my IDE's beautifier.  I'll remember to use the gofmt next time.

PR #211, opened 2018-04-11
Verify PoW for broker node address purchase

EdmundMai:
Verify PoW sent from the webnode

Reference: https://docs.google.com/document/d/1FCS_55lhRd539xl2FEQuIZvzxQDslGhWPsPBEUDLtEk/edit

### DIFF
--- a/actions/transaction_brokernodes.go
+++ b/actions/transaction_brokernodes.go
@@ -1,7 +1,9 @@
 package actions
 
 import (
+	"fmt"
 	"os"
+	"strings"
 
 	"github.com/gobuffalo/buffalo"
 	"github.com/gobuffalo/pop"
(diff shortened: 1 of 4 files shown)


actions/transaction_brokernodes.go
```diff
@@ -74,3 +83,47 @@ func (usr *TransactionBrokernodeResource) Create(c buffalo.Context) error {
...
 	return c.Render(200, r.JSON(res))
 }
+
+func (usr *TransactionBrokernodeResource) Update(c buffalo.Context) error {
+	req := transactionUpdateReq{}
+	parseReqBody(c.Request(), &req)
+
+	// Get transaction
+	t := &models.Transaction{}
+	transactionError := models.DB.Eager("DataMap").Find(t, c.Param("id"))
```
rfornea: Is that supposed to say "DataMap" there?

rfornea: NVM, disregard.

actions/transaction_brokernodes.go
```diff
@@ -74,3 +83,47 @@ func (usr *TransactionBrokernodeResource) Create(c buffalo.Context) error {
...
+	trytes := giota.Trytes(req.Trytes)
+	iotaTransaction, iotaError := giota.NewTransaction(trytes)
+
+	if transactionError != nil || iotaError != nil {
+		return c.Render(400, r.JSON(map[string]string{"error": "No transaction found"}))
+	}
+
+	address, addError := giota.ToAddress(t.DataMap.Address)
+	validAddress := addError == nil && address == iotaTransaction.Address
+	validMessage := giota.Trytes(t.DataMap.Message) == iotaTransaction.SignatureMessageFragment
```
rfornea: This message one might fail sometimes because iota will pad the end of the message with a bunch of 9s if we don't fill it all the way up.

rfornea: Could do something like this:  https://github.com/oysterprotocol/brokernode/blob/master/services/iota_wrappers.go#L405

actions/transaction_brokernodes.go
EdmundMai: yeh it's eager loading that association

PR #214, opened 2018-04-12
Add method to generate a set of indexes for the each sections.

pzhao5:
These indexes are before inserting into the pearl
For example [2, 20, 30] will insert into insert(2), insert(21), insert(32) since each insertion will lead to a shift.

### DIFF
--- a/actions/utils.go
+++ b/actions/utils.go
@@ -7,6 +7,8 @@ import (
 	"math"
 	"math/rand"
 	"net/http"
+	"strconv"
+	"strings"
 )
 
 const (
@@ -74,4 +76,29 @@ func generateInsertedIndexesForPearl(fileSizeInByte int) []int {
 		remainderOfChunks = remainderOfChunks - fileSectorInChunkSize
 	}
 	return indexes
+}
+
+// Convert an int array to a string.
+func IntsJoin(a []int, delim string) string {
+	var buffer bytes.Buffer
+	for i := 0; i < len(a); i++ {
+		buffer.WriteString(strconv.Itoa(a[i]))
+		if i != len(a)-1 {
+			buffer.WriteString(delim)
+		}
+	}
+	return buffer.String()
+}
+
+// Convert an string back to int array
+func IntsSplit(a string, delim string) []int {
+	var ints []int
+	substrings := strings.Split(a, delim)
+	for i := 0; i < len(substrings); i++ {
+		v, e := strconv.Atoi(substrings[i])
+		if e == nil {
+			ints = append(ints, v)
+		}
+	}
+	return ints
 }
--- a/actions/utils.go
+++ b/actions/utils.go
@@ -11,10 +11,10 @@ import (
 
 const (
 	// One chunk unit as represents as 1KB
-	fileChunkSizeInByte = float64(1000)
+	fileChunkSizeInByte = 1000
 
 	// Number of 1KB chunk in one Sector
-	fileSectorInChunkSize = float64(1000000)
+	fileSectorInChunkSize = 1000000
 )
 
 // parseReqBody take a request and parses the body to the target interface.
@@ -65,13 +65,21 @@ func generateInsertedIndexesForPearl(fileSizeInByte int) []int {
 	}
 
 	fileSectorInByte := fileChunkSizeInByte * (fileSectorInChunkSize - 1)
-	numOfSectors := int(math.Ceil(float64(fileSizeInByte) / fileSectorInByte))
-	remainderOfChunks := math.Ceil(float64(fileSizeInByte) / fileChunkSizeInByte)
+	numOfSectors := int(math.Ceil(float64(fileSizeInByte) / float64(fileSectorInByte)))
+	remainderOfChunks := int(math.Ceil(float64(fileSizeInByte) / fileChunkSizeInByte))
 
 	for i := 0; i < numOfSectors; i++ {
-		rang := int(math.Min(fileSectorInChunkSize, remainderOfChunks))
-		indexes = append(indexes, rand.Intn(rang)+i*int(fileSectorInChunkSize))
+		rang := min(fileSectorInChunkSize, remainderOfChunks)
+		indexes = append(indexes, rand.Intn(rang))
 		remainderOfChunks = remainderOfChunks - (fileSectorInChunkSize - 1)
 	}
 	return indexes
+}
+
+// Util method to return min value of a and b.
+func min(a int, b int) int {
+	if a < b {
+		return a
+	}
+	return b
 }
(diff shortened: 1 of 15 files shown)


rfornea:
Everything looks good as far as I can tell, but I do have a question.  

>These indexes are before inserting into the pearl
>For example [2, 20, 30] will insert into insert(2), insert(21), insert(32) since each insertion will lead to a shift.   

So let's pretend that hypothetically, sector size is 100, so if we have 3 sectors they are are 0 through 99, 100 through 199, and 200 through 299.  

Let's pretend that in sector 2 (100-199), the index that we randomly select is 199, however unlikely that would be.  Would the shifting push the index to be 200?  I.e. the sector 2 index winds up being in sector 3?  If it is possible for this to happen, we need to prevent it.

actions/utils.go
```diff
@@ -5,6 +5,19 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	"net/http"
+	"math"
+	"math/rand"
```
AaronVasquez: This is only pseudo random, we should seed it. Probably with a hashed timestamp:

https://golang.org/pkg/hash/
```go
rand.Seed(hash(time.Now().UTC().UnixNano()))
```

actions/utils.go
```diff
@@ -46,3 +59,21 @@ func join(A []string, delim string) string {
...
+// Randomly generate a set of indexes in each sector
+func generateInsertedIndexesForPearl(fileSizeInByte int) []int {
+	if fileSizeInByte <= 0 {
+		return nil
+	}
+
+	numOfSectors := int(math.Ceil(float64(fileSizeInByte) / fileSectorInByte))
+	remainderOfChunks := math.Ceil(float64(fileSizeInByte) / fileChunkSizeInByte)
+
+	var indexes []int
```
AaronVasquez: You can make this fixed length
```go
var indexes [numSectors]int
```

actions/utils.go
```diff
@@ -46,3 +59,21 @@ func join(A []string, delim string) string {
...
 }
+
+// Randomly generate a set of indexes in each sector
+func generateInsertedIndexesForPearl(fileSizeInByte int) []int {
+	if fileSizeInByte <= 0 {
+		return nil
+	}
+
+	numOfSectors := int(math.Ceil(float64(fileSizeInByte) / fileSectorInByte))
+	remainderOfChunks := math.Ceil(float64(fileSizeInByte) / fileChunkSizeInByte)
```
AaronVasquez: I think we should return an array of ints `i` where `0 <= i < sectorChunksCount`, each sector would be able to offset it knowing which sector it's in. So we don't need to know the remainder.

For example, and array `[5, 1]` and sector size of `5` would mean the first treasure is inserted at `3` and second inserted at `11`

We just need to be careful about the last sector, which might not be complete.

actions/utils.go
rfornea: I am not sure but I believe that seeding could be added in main.go, and would be applicable through the application.

actions/utils.go
pzhao5: The problem is that when we insert the pearl, it would push one file chunk to the next sector, thus, all the remaining file chunk will need to shift down.

actions/utils.go
```diff
@@ -5,6 +5,19 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	"net/http"
+	"math"
+	"math/rand"
+)
+
+const (
+	// One chunk unit as represents as 1KB
+	fileChunkSizeInByte = float64(1000)
```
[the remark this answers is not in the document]
pzhao5: I had to change it back to float since i am doing division. Either one are fine, unless you think otherwise

actions/utils.go
```diff
@@ -46,3 +59,21 @@ func join(A []string, delim string) string {
 
 	return buffer.String()
 }
+
+// Randomly generate a set of indexes in each sector
+func generateInsertedIndexesForPearl(fileSizeInByte int) []int {
+	if fileSizeInByte <= 0 {
+		return nil
```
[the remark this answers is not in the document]
pzhao5: Done.

actions/utils.go
pzhao5: Agree, let me look into that. Mind filing an issue for that?

actions/utils.go
pzhao5: That won't work, Because numSectors is not a constant at compile time.

actions/utils.go
AaronVasquez: https://github.com/oysterprotocol/brokernode/issues/217

actions/utils.go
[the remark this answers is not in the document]
AaronVasquez: These will probably be made public later on when used in other calculations. Floats might not make sense there.

actions/utils.go
AaronVasquez: If you have 102 chunks, and 10 chunks per sector, you will know that you will have 11 treasure chunks 
```ceil( 102 / 10 ) = 11```

From this you know there will be 113 total chunks
```102+11```

So your last treasure index will be i where `0 <= i < 3`

So a sample return value is `[ 6, 2, 7, ..., 2 ]`

Comment (pzhao5) [reply actions/utils.go]: Ok, I will remove the offset then

File: actions/utils.go [hunk 6]
```diff
@@ -46,3 +56,30 @@ func join(A []string, delim string) string {
...
+
+	for i := 0; i < numOfSectors; i++ {
+		rang := min(fileSectorInChunkSize, remainderOfChunks)
+		indexes = append(indexes, rand.Intn(rang))
+		remainderOfChunks = remainderOfChunks - (fileSectorInChunkSize - 1)
+	}
+	return indexes
+}
+
+// Util method to return min value of a and b.
```
AaronVasquez: https://golang.org/pkg/math/#Min

actions/utils.go
```diff
@@ -46,3 +56,30 @@ func join(A []string, delim string) string {
...
+	if fileSizeInByte <= 0 {
+		return indexes
+	}
+
+	fileSectorInByte := fileChunkSizeInByte * (fileSectorInChunkSize - 1)
+	numOfSectors := int(math.Ceil(float64(fileSizeInByte) / float64(fileSectorInByte)))
+	remainderOfChunks := int(math.Ceil(float64(fileSizeInByte) / fileChunkSizeInByte))
+
+	for i := 0; i < numOfSectors; i++ {
+		rang := min(fileSectorInChunkSize, remainderOfChunks)
```
AaronVasquez: I think you will have an off by 1 error. You need to add 1 to the remainderOfChunks since the treasure chunk will extend the last sector.

actions/utils.go
pzhao5: Correct, that would only apply for float64. Which mean I need to type cast it into float64 in this case.

AaronVasquez: I see. golang makes me sad sometimes...

actions/utils.go
pzhao5: When I compute the remainderOfChunk, I offset that by 1.

pzhao5: In line 74

actions/utils.go
pzhao5: You are not alone

PR #215, opened 2018-04-12
updating post request with invoice

jfarago:
Updated request and response from client to broker to include PRL invoice and eth address. I created some to do's for code that has not been written yet. I dont know how to use go format, so that might be off.


actions/upload_sessions.go
```diff
@@ -18,15 +18,22 @@ type UploadSessionResource struct {
 // Request Response structs
 
 type uploadSessionCreateReq struct {
-	GenesisHash   string `json:"genesisHash"`
-	FileSizeBytes int    `json:"fileSizeBytes"`
-	BetaIP        string `json:"betaIp"`
+	GenesisHash          string `json:"genesisHash"`
+	FileSizeBytes        int    `json:"fileSizeBytes"`
+	BetaIP               string `json:"betaIp"`
+	StorageLengthInYears int    `json:"storageLengthInYears"`
```
AaronVasquez: Should we store the storage length in the DB too? Could be helpful if we need to debug in the future.

actions/upload_sessions.go
```diff
@@ -161,6 +169,19 @@ func (usr *UploadSessionResource) CreateBeta(c buffalo.Context) error {
...
-	res := uploadSessionCreateRes{UploadSession: u, ID: u.ID.String()}
+	res := uploadSessionCreateRes{
+		UploadSession: u,
+		ID:            u.ID.String(),
+		Invoice:       CreateInvoice(req.StorageLengthInYears, req.FileSizeBytes),
+	}
 	return c.Render(200, r.JSON(res))
 }
+
+func CreateInvoice(storageLengthInYears int, fileSizeBytes int) invoice {
```
AaronVasquez: 1. Let's make `getInvoice` a method on the `uploadSession` model.
    1. This method would return `totalCost` for Cost and `ethAddrAlpha` for EthAddress
2. `calculatePayment` should be called in the `StartUploadSession` method (which would set its own `totalCost` property)

rfornea: @AaronVasquez   
>This method would return totalCost for Cost and ethAddrAlpha for EthAddress   

It seems like we would want to check the type here (for alpha or beta) and then return the alpha or beta address based on that result?

actions/upload_sessions.go
rfornea: I'll add it

actions/upload_sessions.go
AaronVasquez: Yeah, the type should already be set by the time this method gets called. Btw, totalCost should be the entire cost, beta will just expect to receive half. I think this will be less confusing by having upload_sessions cost consistent between the two brokers

PR #218, opened 2018-04-12
Geth integration

AaronVasquez:


services/eth_gateway.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/ethclient"
+)
+
+// TODO: Find etherum node to connect to.
+const ethUrl = "ws://0.0.0.0"
+
+// Singleton client
+var client *ethclient.Client
```
pzhao5: nit:
var (
client *ethclient.Client
mtx sync.Mutex
)

services/eth_gateway.go
```diff
@@ -0,0 +1,67 @@
...
+	if client != nil {
+		return client, nil
+	}
+
+	// check-lock-check pattern to avoid excessive locking.
+	mtx.Lock()
+	defer mtx.Unlock()
+
+	if client != nil {
+		c, err = ethclient.Dial(ethUrl)
```
pzhao5: ethclient is going to be nil in this case.

AaronVasquez: ethclient is the package name (imported on line 11). Unfortunate naming...

services/eth_gateway.go
```diff
@@ -0,0 +1,69 @@
...
+
+// SubscribeToTransfer will subscribe to transfer events
+// sending PRL to the brokerAddr given. Notifications
+// will be sent in the out channel provided.
+func SubscribeToTransfer(brokerAddr common.Address, outCh chan<- types.Log) {
+	ethCl, _ := sharedClient()
+	ctx := context.Background() // TODO: Should we have some timeout or cancel?
+
+	q := ethereum.FilterQuery{
+		FromBlock: nil, // TODO: Figure out how to get current block
```
rfornea: Alex Moyano had to do something similar today for the snapshot script, he probably has some good ideas of how to get the current block

AaronVasquez: @automyr ^

services/eth_gateway.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+
+		// Sets Singleton
+		client = c
+	}
+
+	return client, err
+}
+
+func BuryPrl() {
```
rfornea: Will this method be creating the addresses and getting the private seed keys for the eth addresses we are going to use for the treasure?

services/eth_gateway.go
automyr: More concretely, here's the method using the JSON RPC API:

https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_blocknumber

services/eth_gateway.go
AaronVasquez: I thought this was a function in the smart contract called from the eth address of the upload session?

I will have to add another function to generate a new eth address for an upload session though.

rfornea: There's definitely a bury function but I'm not sure that it creates those addresses.

PR #221, opened 2018-04-13
Method to generate ethereum addresses and private key

AaronVasquez:


actions/upload_sessions.go
```diff
@@ -19,14 +21,22 @@ type UploadSessionResource struct {
...
-	FileSizeBytes        int    `json:"fileSizeBytes"`
-	BetaIP               string `json:"betaIp"`
-	StorageLengthInYears int    `json:"storageLengthInYears"`
-	AlphaTreasureIndexes []int  `json:"alphaTreasureIndexes"`
+	GenesisHash          string         `json:"genesisHash"`
+	FileSizeBytes        int            `json:"fileSizeBytes"`
+	BetaIP               string         `json:"betaIp"`
+	StorageLengthInYears int            `json:"storageLengthInYears"`
+	AlphaTreasureIndexes []int          `json:"alphaTreasureIndexes"`
+	Invoice              models.Invoice `json:"invoice"`
```
rfornea: Maybe I am missing something obvious, but what do we need Invoice on the request for?  I see you're using it on lines 195 and 196, but couldn't you just use the invoice you made on line 80?

models/upload_sessions.go
```diff
@@ -2,14 +2,15 @@ package models
```
rfornea: There's some commented out lines in this file that we could remove but that's no biggie.

actions/upload_sessions.go
AaronVasquez: Lines 195 and 196 are in a different handler (`/upload-sessions/beta`), so alpha needs to pass the cost and it's ETH address to beta. 

We are reusing the same request for alpha and beta, which I agree is messy though. I can copy and paste it, unless there's a cleaner way to share fields in a struct in golang?

rfornea: Ah I see.  I'm sure there's a better way to share fields in structs but I don't think figuring that out needs to be a priority.

PR #222, opened 2018-04-13
Add init impl for Merge int[] for treasure map

pzhao5:


actions/upload_sessions.go
```diff
@@ -26,11 +27,11 @@ type uploadSessionCreateReq struct {
...
-	ID              string               `json:"id"`
-	UploadSession   models.UploadSession `json:"uploadSession"`
-	BetaSessionID   string               `json:"betaSessionId"`
-	Invoice         models.Invoice       `json:"invoice"`
-	BetaBuriedIndex []int                `json:"betaBuriedIndexes"`
+	ID                string               `json:"id"`
+	UploadSession     models.UploadSession `json:"uploadSession"`
+	BetaSessionID     string               `json:"betaSessionId"`
+	Invoice           models.Invoice       `json:"invoice"`
+	BetaBuriedIndexes []int                `json:"betaBuriedIndexes"`
```
AaronVasquez: Nit: These aren't buried yet, lets rename it to BetaTreasureIndex (same with alpha above)

actions/utils.go
```diff
@@ -60,24 +63,38 @@ func join(A []string, delim string) string {
...
+		rang := int(math.Min(FileSectorInChunkSize, remainderOfChunks))
 		indexes = append(indexes, rand.Intn(rang))
-		remainderOfChunks = remainderOfChunks - fileSectorInChunkSize
+		remainderOfChunks = remainderOfChunks - FileSectorInChunkSize
 	}
 	return indexes
 }
 
+// Merge 2 different indexes into 1 indexes. Computed Merged indexes
+func MergeIndexes(a []int, b []int) []int {
```
AaronVasquez: This should be an error if the lengths don't match.

PR #231, opened 2018-04-16
Add Transform Indexes method for Pearl bury method

pzhao5:


utils/utils.go
```diff
@@ -55,6 +55,21 @@ func ParseResBody(res *http.Response, dest interface{}) (err error) {
 	return
 }
 
+// Transforms index with correct position for insertion after considering the buried indexes.
+func TransformIndexWithBuriedIndexes(index int, treasureIdxMap []int) int {
+	if len(treasureIdxMap) == 0 {
```
AaronVasquez: There should always be treasure. This should return an error. Also make it `<= 0` to catch more error cases.

pzhao5: In that case, the client would need to handle this. And client would still need to decide how to handle the error.
1) Just crash the app.
2) Skip burying part
3) Just default to original index. I think a logging here would be nice rather than throwing error.

Let me know what you think

AaronVasquez: That's true. We should log here since its an invalid state. I don't think we should crash the app (this is not elixir unfortunately) since this error is independent from the other datamaps and we can't isolate the panic. The client should be notified of this error because something went wrong.

With that said, I don't think this state will happen, so just a TODO is fine for now.

PR #234, opened 2018-04-17
Process paid sessions 2

rfornea:
-Expects a TreasureIdxMap that looks like this:
`[{
		"sector": 1,
		"idx": 155,
		"key": "firstKeySecondMap"
		},
		{ 
		"sector": 2,
		"idx": 204,
		"key": "secondKeySecondMap"
		},
		{
		"sector": 3,
		"idx": 599,
		"key": "thirdKeySecondMap"
		}]`
-Checks for sessions which have been paid but in which treasure has not been buried, and buries the treasure
-Checks for sessions which have been paid and have treasure buried and marks them as Unassigned so that process_unassigned_chunks will pick them up
-Deletes keys when we are done with them
-Unit tests for all of the above

Wound up writing the method in a way that didn't rely on "Where In" working correctly, wish I'd done that from the start.

### DIFF
--- a/jobs/process_paid_sessions.go
+++ b/jobs/process_paid_sessions.go
@@ -2,10 +2,17 @@ package jobs
 
 import (
 	"encoding/json"
+	"errors"
 	"github.com/getsentry/raven-go"
-	"github.com/gobuffalo/pop/slices"
 	"github.com/oysterprotocol/brokernode/models"
+	"log"
 )
+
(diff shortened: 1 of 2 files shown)


rfornea:
We will need to revisit this when we work on the task to encrypt and decrypt the keys.  We'll need to encrypt the keys in process_paid_sessions_test.go and decrypt in process_paid_sessions.go

PR #237, opened 2018-04-18
Added get response for upload_sessions and tests

jfarago:
Currently, when the ID does not exist we return the error straight from the database. I was having trouble wiring up a custom, more user friendly, error. @rfornea could probably elaborate better than I could on this.

### DIFF
--- a/actions/upload_sessions.go
+++ b/actions/upload_sessions.go
@@ -54,6 +54,11 @@ type UploadSessionUpdateReq struct {
 	Chunks []chunkReq `json:"chunks"`
 }
 
+type paymentStatusCreateRes struct {
+	ID            string               `json:"id"`
+	PaymentStatus string			   `json:"paymentStatus"`
+}
+
 // Create creates an upload session.
 func (usr *UploadSessionResource) Create(c buffalo.Context) error {
 	req := uploadSessionCreateReq{}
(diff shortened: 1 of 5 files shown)


actions/upload_sessions.go
```diff
@@ -215,3 +220,31 @@ func (usr *UploadSessionResource) CreateBeta(c buffalo.Context) error {
...
+	err := models.DB.Where("id = ?", c.Param("id")).First(&session)
+
+	if err != nil {
+		//TODO: Return better error response when ID does not exist
+		return err
+	}
+
+	paymentStatus := ""
+
+	if session.PaymentStatus == models.PaymentStatusPending {
```
AaronVasquez: Nit: a `switch` statement would be a little cleaner. Also can you make this a method in the model file for `upload_sessions`?

actions/upload_sessions.go
```diff
@@ -215,3 +220,31 @@ func (usr *UploadSessionResource) CreateBeta(c buffalo.Context) error {
 	}
 	return c.Render(200, r.JSON(res))
 }
+
+
+func (usr *UploadSessionResource) GetPaymentStatus(c buffalo.Context) error {
+	session := models.UploadSession{}
+	err := models.DB.Where("id = ?", c.Param("id")).First(&session)
+
+	if err != nil {
```
AaronVasquez: Does pop return this as an err or would `session` be nil?

actions/upload_sessions.go
```diff
@@ -215,3 +220,31 @@ func (usr *UploadSessionResource) CreateBeta(c buffalo.Context) error {
...
+		//TODO: Return better error response when ID does not exist
+		return err
```
AaronVasquez: render a `404`

actions/upload_sessions.go
rfornea: We actually ran into json syntax errors when trying to render something to return.  

We need to render two types of errors I suppose:
-when something is actually wrong
-when nothing is wrong but there's just no session matching that ID (we were going to use 204 for this)  

@jfarago You could take a look at some of the other files in the actions folder, particularly for methods that are handlers for URLs that end with {id}.  There may already be an example of returning an error when no such session exists.

AaronVasquez: we can hold off on rendering better errors for now. since we are the only client, we know how to handle it.

However when something is not found, will pop give you an error or an empty struct? If an empty struct, we should check `if session == nil || err != nil`

rfornea: I am not entirely sure if it gives you an empty struct in this case.  I think it may just give you a struct in which all fields have been assigned to default values, such as "" for strings and 0 for ints, etc.  

I often use .All and then check for len(arrayOfThings) == 0 for this reason, but another option would be to still use .First and check that session.ID == (whatever the default value for a uuid would be)

actions/upload_sessions.go
```diff
@@ -215,3 +220,21 @@ func (usr *UploadSessionResource) CreateBeta(c buffalo.Context) error {
 	}
 	return c.Render(200, r.JSON(res))
 }
+
+
+func (usr *UploadSessionResource) GetPaymentStatus(c buffalo.Context) error {
+	session := models.UploadSession{}
+	err := models.DB.Find(&session, c.Param("id"))
+
+	if err != nil || session.ID == models.EmptyUUID {
```
AaronVasquez: I think you can do `session == models.UploadSession{}` or `session.ID == uuid.Nil`.

jfarago: Sounds good, I chose the former so we dont have to import uuid

PR #239, opened 2018-04-19
Create math to convert trybyte to byte and vice verse

pzhao5:


utils/utils.go
```diff
@@ -14,6 +14,12 @@ import (
 )
 
 const (
+	// 1000 Byte or 1KB
+	KiloByte = 1000
+
+	// 1000 Byte (1KB) = 2187 Trytes
+	KiloByteToTrytes = 2187
```
rfornea: Technically this should be 2000.  

1000 bytes in ascii, upon being converted to trytes, will take up 2000 bytes of space.  

2187 is just the maximum number of single characters that can be stuffed into a "message".  Since we receive the "fileSizeBytes" after it has already been converted to trytes, the value we are receiving is the final amount of space that we will be taking up.  So say we receive a file size of 8748 bytes.  This is (2187 / 4).  So we would know that we must build a data map with 4 chunks + a metadata chunk, in other words 5 chunks.  

The original size of the data before being encrypted and converted was probably something like ~4000 bytes or whatever, but that doesn't really matter.  The broker receives it after it's already been tryted.

rfornea: To clarify, we definitely still need to be using 2187, but when I said "Technically this should be 2000." in reference to "KiloByteToTrytes" I was referring to the fact that 1000 bytes of regular data will take up 2000 bytes of space upon being converted to trytes.  We still want to use 2187, I just think we might should name it something different like "MaxMessageSize" or something, to be less confusing.

PR #245, opened 2018-04-24
Claim unused prls

rfornea:
-adds completed_uploads table
-adds claim_unused_prls task
-adds the ability to mock methods from the eth_wrapper.go file.  Currently I've only mocked the two methods I need to use.  Additional "generic" mocks can be added to eth_wrapper_mock.go, and then in your tests, re-assign individual methods to more specific mocks.


jobs/claim_unused_prls.go
```diff
@@ -0,0 +1,147 @@
...
+
+	SendGasForNewClaims()
+	StartNewClaims()
+
+	PurgeCompletedClaims()
+}
+
+func SetEthWrapper(ethService services.Eth) {
+	ethWrapper = ethService
+}
```
rfornea: Using this to tell this file to use the EthMock instead of the real thing, for the tests.

jobs/claim_unused_prls_test.go
```diff
@@ -0,0 +1,295 @@
+package jobs_test
+
+import (
+	"github.com/oysterprotocol/brokernode/jobs"
+	"github.com/oysterprotocol/brokernode/models"
+	"github.com/oysterprotocol/brokernode/services"
+	"time"
+)
```
rfornea: Structured the tests in these two test files a little differently since I'm re-using the same test data throughout the file.

jobs/job_runner.go
```diff
@@ -162,3 +173,17 @@ var processPaidSessionsHandler = func(args worker.Args) error {
 
 	return nil
 }
+
+var claimUnusedPRLsHandler = func(args worker.Args) error {
+	thresholdTime := time.Now().Add(-3 * time.Hour) // consider a transaction timed out if it takes more than 3 hours
```
rfornea: We can change this timed out time (3 hours) if we want but I know sometimes it takes a long time if the network is congested.

jobs/flush_old_webnodes_test.go
```diff
@@ -16,13 +16,13 @@ func (suite *JobsSuite) Test_FlushOldWebnodes() {
 	// setting the threshold time to be more recent than when the webnodes were updated
 	thresholdTime := time.Now().Add(time.Millisecond * 10000)
 
-	suite.Equal(err, nil)
+	suite.Equal(nil, err)
```
rfornea: These are all nits, first param is supposed to be the expected value.

models/completed_uploads.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func GetTimedOutPRLTransfers(thresholdTime time.Time) (uploads []CompletedUpload, err error) {
+	err = DB.Where("prl_status = ? AND updated_at <= ?",
+		PRLClaimProcessing,
+		thresholdTime).All(&uploads)
+	return uploads, err
+}
+
+func SetGasStatusByAddress(transactionAddress string, newGasStatus int) {
```
rfornea: The intention with this method and the one below is for the eth_gateway file to all this when it hears a successful transfer event, to change the GasStatus or PRLStatus

services/eth_gateway.go
```diff
@@ -13,17 +18,70 @@ import (
...
+	// Load ENV variables
+	err := godotenv.Load()
+	if err != nil {
+		log.Println("Error loading .env file")
+		raven.CaptureError(err, nil)
+	}
+
+	MainWalletAddress := os.Getenv("MAIN_WALLET_ADDRESS")
+	MainWalletKey := os.Getenv("MAIN_WALLET_KEY")
+	ethUrl := os.Getenv("ETH_NODE_URL")
```
rfornea: Put all the goodies in the .env file.  The print statements below are because I was getting unused variable errors.

services/eth_gateway.go
```diff
@@ -13,17 +18,70 @@ import (
...
+	ClaimPRLs           ClaimPRLs
+	GenerateEthAddr     GenerateEthAddr
+	BuryPrl             BuryPrl
+	SendETH             SendETH
+	SendPRL             SendPRL
+	GetGasPrice         GetGasPrice
+	SubscribeToTransfer SubscribeToTransfer
+	CheckBalance        CheckBalance
+	GetCurrentBlock     GetCurrentBlock
+}
```
rfornea: Changed this file some to allow for easier mocking.

services/eth_gateway_mock.go
```diff
@@ -0,0 +1,23 @@
...
+
+func SetUpMock() {
+
+	EthMock = Eth{
+		ClaimPRLs: func([]models.CompletedUpload) error {
+			return nil
+		},
+		SendGas: func([]models.CompletedUpload) error {
+			return nil
+		},
```
rfornea: Add generic mocks for other methods here, once we know for sure the params and return values.

rfornea:
The travis build didn't run.  :/  Anybody know why?

services/eth_gateway.go
AaronVasquez: Can you document all the ENV vars needed to make sure we don't miss any here:
https://docs.google.com/document/d/1B7rDoOwWV7yPgJY4FxD0DvynvjQIv4BdGOQbjIP84cI/edit

models/completed_uploads.go
AaronVasquez: Instead of this API taking in an int, we can add a type to constrain it:

```go
type GasTransferStatus int
const (
	GasTransferNotStarted GasTransferStatus = iota + 1
	GasTransferProcessing
	GasTransferSuccess
	GasTransferError = -1
)
// ...
func SetGasStatusByAddress(transactionAddress string, status GasTransferStatus) 
```

services/eth_gateway.go
rfornea: done

models/completed_uploads.go
rfornea: sure

PR #249, opened 2018-04-26
Create a batch update to data_maps table

pzhao5:
Have a better way to convert a struct{} data with db tag into string format for update a table

### DIFF
--- a/utils/db_utils.go
+++ b/utils/db_utils.go
@@ -14,14 +14,19 @@ import (
 
 type ValueT interface{}
 
+// Interface for DB Update operation.
 type dbUpdateOperation interface {
+	// Get columns in string format and separated by ", "
 	GetColumns() string
+	// Get columns new value in string format and separated by ", "
 	GetNewUpdateValue(ValueT) string
 }
 
+const COLUMNS_SEPARATOR = ", "
+
+// Internal data structure
 type dbUpdateModel struct {
 	columns        columns.Columns
-	primayeKeyType string
 	fieldMap       map[string]string // Map from tableColumn to fieldName
 }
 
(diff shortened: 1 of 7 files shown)


pzhao5:
Sure, how about I do a follow up on this issue. Added a TODO here

utils/db_utils.go
```diff
@@ -0,0 +1,111 @@
...
+	"fmt"
+	"github.com/gobuffalo/pop"
+	"github.com/gobuffalo/pop/columns"
+	"github.com/gobuffalo/uuid"
+	"github.com/pkg/errors"
+	"reflect"
+	"sort"
+	"strconv"
+	"strings"
+)
```
rfornea: A lot of the variables in this file could use more descriptive names, to make it easier for other developers to be able to quickly look at know what the code is doing.  No need to worry about it for this PR, could address if when you push the PR to do the batch updates every X iterations.

pzhao5: Can you give me an example? I could update it.

rfornea: Yeah.  
c := columns.ColumnsForStructWithAlias(vPtr, m.TableName(), m.As)   
Could be something like:
columns := columns.ColumnsForStructWithAlias(vPtr, model.TableName(), model.As)

And something like this:  
m := pop.Model{Value: vPtr}
Could be something like this:
model := pop.Model{Value: vPtr}

Or like this:
xs = append(xs, fmt.Sprintf("'%s'", u.String()))  
It's not super obvious what it is.  I'm guessing those are strings with the column names, so maybe something like "columnName" or something.

pzhao5:
I see. Will update it

PR #252, opened 2018-04-27
Qa whole flow

rfornea:
This PR was getting really huge so I wanted to go ahead and commit what I've got so far, and I'll address some more things in another PR in the next day or so.  

This PR adds the following:
-in the .env file, lets you specify which "mode" you want the broker to run in depending on your needs.  PROD_MODE, TEST_MODE_NO_TREASURE, or TEST_MODE_DUMMY_TREASURE.  Default is PROD_MODE and the unit tests run in PROD_MODE, which means I had to add 1 to many of the unit tests for every unique data map, for the treasure chunks.
-determines treasure chunks with numChunks instead of file size.  We probably should have done this a long time ago, not sure why we ever used file size for this to begin with.  Still currently need to send file size to calculate payment.  In a future PR I can work on pulling this out, but this should be a low priority IMO.  
-deletes .env file, adds it to .gitignore, adds .env.example, and adds commands to travis.yml so copy the .env.example to .env, so we don't have to check the .env file in anymore, yay!  
-When running in TEST_MODE_NO_TREASURE, we do the following things differently:
- automatically set chunks to "unassigned" when their message comes in
- don't make treasure offsets
- automatically set all sessions to "PAID" and "BURIED"
- don't run claim_unused_prls task        

In my next round of brokernode changes I'll be working on the TEST_MODE_DUMMY_TREASURE and hopefully QAing that the webnode is able to find the "dummy" treasure.  When Astor is done with the Ethereum stuff we can QA PROD_MODE.


models/data_maps.go
```diff
@@ -283,3 +288,12 @@ func insertsIntoDataMapsTable(columnsName string, values string) error {
 	rawQuery := fmt.Sprintf("INSERT INTO %s (%s) VALUES %s", DataMapTableName, columnsName, values)
 	return DB.RawQuery(rawQuery).All(&[]DataMap{})
 }
+
+func GetDataMapByGenesisHashAndChunkIdx(genesisHash string, chunkIdx int) (dataMaps []DataMap, err error) {
```
pzhao5: Please add a comment on this method.

models/data_maps.go
```diff
@@ -283,3 +288,12 @@ func insertsIntoDataMapsTable(columnsName string, values string) error {
...
+	dataMaps = []DataMap{}
+	err = DB.Where("genesis_hash = ?",
+		genesisHash).Where("chunk_idx = ?", chunkIdx).All(&dataMaps)
+
+	return dataMaps, err
```
pzhao5: nit: just do: return
since you already have dataMaps and err as declaration. 
Also, it is more readability do this:

func GetDataMapByGenesisHashAndChunkIdx(genesisHash string, chunkIdx int) ([]DataMap, error) {

models/upload_sessions.go
```diff
@@ -236,3 +271,21 @@ func GetSessionsByAge() ([]UploadSession, error) {
 
 	return sessionsByAge, nil
 }
+
+func GetSessionsThatNeedTreasure() ([]UploadSession, error) {
```
pzhao5: add method document

utils/mode.go
```diff
@@ -0,0 +1,50 @@
...
+		raven.CaptureError(err, nil)
+	}
+
+	brokerMode := os.Getenv("MODE")
+
+	SetBrokerMode(brokerMode)
+}
+
+func SetBrokerMode(brokerMode string) {
+
```
pzhao5: nit: remove nl

utils/mode.go
```diff
@@ -0,0 +1,50 @@
...
+	// Load ENV variables
+	err := godotenv.Load()
+	if err != nil {
+		log.Println("Error loading .env file")
+		raven.CaptureError(err, nil)
+	}
+
+	brokerMode := os.Getenv("MODE")
+
+	SetBrokerMode(brokerMode)
```
pzhao5: this is a private method? maybe do: setBrokerMode()

models/data_maps.go
rfornea: Okay I'll change this.  Since I'm getting rid of the declared return values I'll still have to explicitly return dataMaps and err.

utils/mode.go
rfornea: Okay I'll make it private for now.  We may have to re-expose it if we decide to do any testing of the different broker modes.

PR #260, opened 2018-05-03
treasure hunting 

tenisakb:
build datamap
prepared cycle for checking IOTA 
checking epoch

### DIFF
--- a/actions/treasures.go
+++ b/actions/treasures.go
@@ -1,6 +1,15 @@
 package actions
 
-import "github.com/gobuffalo/buffalo"
+import (
+	"fmt"
+	"os"
+	"time"
+	"strings"
+	"github.com/gobuffalo/buffalo"
+	"github.com/iotaledger/giota"
+	"github.com/oysterprotocol/brokernode/models"
+	"github.com/oysterprotocol/brokernode/utils"
+)
 
 type TreasuresResource struct {
 	buffalo.Resource
@@ -15,7 +24,51 @@ type treasureReq struct {
 	EthKey          string `json:"EthKey"`
 }
 
+type treasureRes struct {
+	Success string `json:"success"`
+}
+
+func unixMilli(t time.Time) int64 {
+	return t.Round(time.Millisecond).UnixNano() / (int64(time.Millisecond) / int64(time.Nanosecond))
+}
+
 // Verifies the treasure and claims such treasure.
 func (t *TreasuresResource) VerifyAndClaim(c buffalo.Context) error {
-	return nil
+	req := treasureReq{}
+	oyster_utils.ParseReqBody(c.Request(), &req)
+
+	host_ip := os.Getenv("HOST_IP")
+	provider := "http://" + host_ip + ":14265"
+	api := giota.NewAPI(provider, nil)
+	ftr := &giota.FindTransactionsRequest{Bundles: []giota.Trytes{"DEXRPLKGBROUQMKCLMRPG9HFKCACDZ9AB9HOJQWERTYWERJNOYLW9PKLOGDUPC9DLGSUH9UHSKJOASJRU"}}
+	resp, err := api.FindTransactions(ftr)
+
+	datamap1, vErr := models.GetDataMap(req.GenesisHash, req.NumChunks)
+	for _, d := range datamap1 {
+		
+	}
+	// or 
+	
+	vErr, error := models.BuildDataMaps(req.GenesisHash, req.NumChunks)
+	datamap, err := models.GetDataMapByGenesisHashAndChunkIdx(req.GenesisHash, req.NumChunks)
+	for _, d := range datamap {
+		
+	}
+	
+	var transactions[2] int64
+	transactions[0] = 11
+	transactions[0] = 1122
+	epoch := unixMilli(time.Now().AddDate(-1, 0, 0))
+	verifyAttached := true
+	for _, transaction := range transactions {
+		if transaction <= epoch {
+			verifyAttached = false
+		}
+	}
+
+	res := treasureRes{
+		Success:       "true"
+	}
+
+	return c.Render(202, r.JSON(res))
 }
--- a/models/data_maps.go
+++ b/models/data_maps.go
@@ -298,3 +298,30 @@ func GetDataMapByGenesisHashAndChunkIdx(genesisHash string, chunkIdx int) ([]Dat
 
 	return dataMaps, err
 }
+
+func GetDataMap(genHash string, numChunks int) (dataMap []DataMap, vErr *validate.Errors) {
+
+	fileChunksCount := numChunks
+
+	if oyster_utils.BrokerMode != oyster_utils.TestModeNoTreasure {
+		fileChunksCount = oyster_utils.GetTotalFileChunkIncludingBuriedPearlsUsingNumChunks(numChunks)
+	}
+
+	currHash := genHash
+	for i := 0; i < fileChunksCount; i++ {
+		obfuscatedHash := oyster_utils.HashString(currHash, sha512.New384())
+		currAddr := string(oyster_utils.MakeAddress(obfuscatedHash))
+
+		dataMap := DataMap{
+			GenesisHash:    genHash,
+			ChunkIdx:       i,
+			Hash:           currHash,
+			ObfuscatedHash: obfuscatedHash,
+			Address:        currAddr,
+			Status:         Pending,
+		}
+		// Validate the data
+		vErr, _ = dataMap.Validate(nil)
+	}
+	return
+}
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ data/
 migrations/schema.sql
 .env
 *.env
+.DS_Store


actions/treasures.go
```diff
@@ -15,7 +24,51 @@ type treasureReq struct {
...
+	ftr := &giota.FindTransactionsRequest{Bundles: []giota.Trytes{"DEXRPLKGBROUQMKCLMRPG9HFKCACDZ9AB9HOJQWERTYWERJNOYLW9PKLOGDUPC9DLGSUH9UHSKJOASJRU"}}
+	resp, err := api.FindTransactions(ftr)
+
+	datamap1, vErr := models.GetDataMap(req.GenesisHash, req.NumChunks)
+	for _, d := range datamap1 {
+		
+	}
+	// or 
+	
+	vErr, error := models.BuildDataMaps(req.GenesisHash, req.NumChunks)
```
rfornea: Calling this method will actually insert the datamap into the brokernode's DB table.

actions/treasures.go
```diff
@@ -15,7 +24,51 @@ type treasureReq struct {
...
+	datamap, err := models.GetDataMapByGenesisHashAndChunkIdx(req.GenesisHash, req.NumChunks)
```
rfornea: This method queries the DB data_maps table.  The data_map that the brokernode builds for claiming the treasure shouldn't end up in the data_maps table.

actions/treasures.go
```diff
@@ -15,7 +24,51 @@ type treasureReq struct {
...
+	for _, d := range datamap {
+		
+	}
```
rfornea: Is something supposed to be happening in that ^ for loop?

actions/treasures.go
```diff
@@ -15,7 +24,51 @@ type treasureReq struct {
...
+	
+	var transactions[2] int64
+	transactions[0] = 11
+	transactions[0] = 1122
```
rfornea: What are these for?  It looks like you're setting transactions[0] then immediately setting it to something else?

models/data_maps.go
```diff
@@ -298,3 +298,30 @@ func GetDataMapByGenesisHashAndChunkIdx(genesisHash string, chunkIdx int) ([]Dat
 
 	return dataMaps, err
 }
+
+func GetDataMap(genHash string, numChunks int) (dataMap []DataMap, vErr *validate.Errors) {
+
+	fileChunksCount := numChunks
+
+	if oyster_utils.BrokerMode != oyster_utils.TestModeNoTreasure {
```
rfornea: It looks like what you're doing here is building a datamap on the fly and not saving it to the DB?  That makes sense, though we might want to insert it (into a different table than data_maps) depending on how long the verification takes.  

Either way, you'll want to remove lines 306 through 308.  Those "modes" are for attaching new data_maps, not verifying old ones.

models/data_maps.go
```diff
@@ -298,3 +298,30 @@ func GetDataMapByGenesisHashAndChunkIdx(genesisHash string, chunkIdx int) ([]Dat
...
+			GenesisHash:    genHash,
+			ChunkIdx:       i,
+			Hash:           currHash,
+			ObfuscatedHash: obfuscatedHash,
+			Address:        currAddr,
+			Status:         Pending,
+		}
+		// Validate the data
+		vErr, _ = dataMap.Validate(nil)
+	}
```
rfornea: Maybe i'm reading this wrong but it looks like what would get returned from this method is just a single data_map (whatever the value of the last one is), not an array of data_maps.  I'd think you'd need to append all those data_maps to an array.

actions/treasures.go
```diff
@@ -15,7 +24,51 @@ type treasureReq struct {
...
+}
+
 // Verifies the treasure and claims such treasure.
 func (t *TreasuresResource) VerifyAndClaim(c buffalo.Context) error {
-	return nil
-}
+	req := treasureReq{}
+	oyster_utils.ParseReqBody(c.Request(), &req)
+
+	host_ip := os.Getenv("HOST_IP")
```
rfornea: I think it would be better to pull the logic from lines 40-44 into a new method in the services/iota_wrappers file.  Lines 40-42 could even be omitted since that iota_wrapper file already has the API instantiated.

actions/treasures.go
```diff
@@ -15,7 +24,51 @@ type treasureReq struct {
...
+	provider := "http://" + host_ip + ":14265"
+	api := giota.NewAPI(provider, nil)
+	ftr := &giota.FindTransactionsRequest{Bundles: []giota.Trytes{"DEXRPLKGBROUQMKCLMRPG9HFKCACDZ9AB9HOJQWERTYWERJNOYLW9PKLOGDUPC9DLGSUH9UHSKJOASJRU"}}
```
rfornea: I don't think you will be looking for Bundles.  You'll be looking for addresses.

actions/treasures.go
```diff
@@ -15,7 +24,51 @@ type treasureReq struct {
...
+	var transactions[2] int64
+	transactions[0] = 11
+	transactions[0] = 1122
+	epoch := unixMilli(time.Now().AddDate(-1, 0, 0))
+	verifyAttached := true
+	for _, transaction := range transactions {
+		if transaction <= epoch {
+			verifyAttached = false
+		}
+	}
```
rfornea: Just to make sure I'm understanding what is going on here -- the transactions variable is just a "dummy" variable for now, and you'll replace that with the transaction times of the actual transactions in a future PR?

tenisakb: yes, in the time when I am getting transaction.timestamp that is compare with epoch

actions/treasures.go
tenisakb: Is FindTransactionsRequest correct? How it should looks? Should it be like Address: [].giota.toAddress{req.GenesisHash}  ?

How can I transform result from FindTransactionsRequest to datamap or array (then I would like to walk through)

actions/treasures.go
tenisakb: Yes, so where to save datamap? To database or on fly or dont save? If into database, should I create new table or is something prepared?

actions/treasures.go
tenisakb: Ok, so the same like previous comment... should I save it into db or on fly? Or dont save?

actions/treasures.go
tenisakb: There will be walk throug of datamap and validation and probably validation what I still dont have

actions/treasures.go
tenisakb: Its dummy transaction

models/data_maps.go
tenisakb: Ok so we will ONLY save into db? Any idea about name of table?

So I will remove lines 306 through 308 + save data to db?

actions/treasures.go
tenisakb: FindTransactionsRequest is correct for searching transactionsObjects? 
Commad: "findTransactions" or Command:   "findTransactionsObjects”?

If I dont want to search by bundles how it should looks?

Is this example correct for validation with timestamp?

chunks []models.DataMap
addresses := make([]giota.Address, 0, len(chunks))
for _, chunk := range chunks {
	if chunk.Status != models.Error {
		addresses = append(addresses, giota.Address(chunk.Address))
	}
}
request := giota.FindTransactionsRequest{
	Command:   "findTransactions", 
	Addresses: addresses,
}
response, err := api.FindTransactions(&request)

trytesArray, err := api.GetTrytes(response.Hashes)

transactionObjects := map[giota.Address][]giota.Transaction{}

for _, txObject := range trytesArray.Trytes {
	 transactionObjects[txObject.Address].timestamp  — timestamp???
}

actions/treasures.go
tenisakb: Should I inicializate iota_wrapper - function ini in iota_wrappers? If yes, how? Or is enaugh only import? Function should I insert into structure IotaService in iota_wrappers - example: VerifyTreasure(Datamap, bool) ?

models/data_maps.go
tenisakb: I think this comment will resolve it: https://github.com/oysterprotocol/brokernode/pull/260#discussion_r186021976

actions/treasures.go
rfornea: There is currently no table yet for saving data maps that the brokernode is currently in the process of doing PoW for.  So, if we decide to go that route, you'll have to create one.  

There's some decisions we have to make.  We probably don't want the brokernode to have to verify that ALL transactions in the sector had PoW done on them within the last epoch, since that could be up to 1 million verifications.  What I think we should do is have the brokernode randomly verify that X transactions from the sector have had the PoW done in the last epoch, where X is some number that's large enough that the webnode will not feel comfortable cheating since it will likely get caught.  

I would recommend we build this in such a way that X is easily changed, and divide the sector to be verified into X subsectors.  Each subsector would have a length of 1,000,000 / X.  Randomly pick a single transaction from within each subsector to verify.  That way we aren't re-verifying any transactions we've already verified and we know that our verifications are evenly distributed throughout the datamap.  

Whether it makes sense to save this to a DB table or do it on the fly really depends on how large X is.  If X is fairly small, doing it on the fly would be fine.  If X is fairly large, I would think we'd want it in a table.  

I'm leaning towards suggesting we need tables for this.  Doing it in a table would still work, even if X is small, whereas doing it on the fly might not work so well if X is large, so with a table we've got both our bases covered.  You'd probably want 2 tables.  One named something like "verification_sessions" which would store the values sent to it by the webnode and anything else it needs, including the payout ETH address, and then another called something like "verifiable_data_maps" that would store the session ID of the verification request and the actual addresses and the timestamps of the transactions.  

With mainnet coming up soon though, we could always just keep X fairly small, and do the checking on the fly, and then implement a change to move it to a table after mainnet for more thorough verification.  

@AaronVasquez    thoughts?

actions/treasures.go
rfornea: ok

actions/treasures.go
rfornea: Gotcha

models/data_maps.go
rfornea: Remove lines 306 through 308.  Let's discuss in slack whether we want the data in a table or just to do it on the fly.

actions/treasures.go
rfornea: In the iota_wrappers file, you'll add a (lowercase) method called `verifyTreasure`.  

In the iota_wrappers file, in the struct on lines 98-103, you'll add:  
`VerifyTreasure:  verifyTreasure`
https://github.com/oysterprotocol/brokernode/blob/master/services/iota_wrappers.go#L98

Beneath the types declared in lines 42-45 in iota_wrappers, you'll add a type declaration:
`type VerifyTreasure func(types of the parameters you are passing in) (types of the return values)`
https://github.com/oysterprotocol/brokernode/blob/master/services/iota_wrappers.go#L42

In the struct on line 35-40, you'll add the following:
`VerifyTreasure   VerifyTreasure`   
https://github.com/oysterprotocol/brokernode/blob/master/services/iota_wrappers.go#L35

When you want to use it in your file, you'll import "github.com/oysterprotocol/brokernode/services" and then in your variables, you'll add this:  
`var IotaWrapper = services.IotaWrapper` 
Here is an example of this in the job_runner.go file:  
https://github.com/oysterprotocol/brokernode/blob/master/jobs/job_runner.go#L13

Then IotaWrapper will be available in your file and you can do `IotaWrapper.MethodIWantToCall()` etc.

actions/treasures.go
AaronVasquez: Agree with everything you just said. Lets go with on the fly for this PR, and work on optimizing and fault tolerance later. Let's start with verifying 5 random transactions? This number is probably too small, but we can tweak this in the future.

actions/treasures.go
rfornea: Here's how to build the FindTransactionRequest:
https://godoc.org/github.com/iotaledger/iota.lib.go#FindTransactionsRequest   

As for the address, you'll get those when you build the dataMap.  

It'll be something like this:  
```
sectorDataMap := BuildSectorDataMap(genesisHash, sector, numChunks)
   /* BuildSectorDataMap will be some new method you write that does not save to the data_maps table
Only need to build the dataMap up until the end of the sector the webnode did.  Need numChunks to know when to stop building the dataMap, in the event that the sector the webnode passed was the last sector.  I.e. if there are only 2,500,000 hashes in the dataMap and the webnode did Sector 3, need the numChunks so we know to stop at 2,500,000.   

BuildSectorDataMap would not need to do the obfuscation and tryting for the whole map, just the section that you care about.  You'll still have to get the sha256 of all the chunks leading up to the sector you care about.  So let's say it's sector 3 in a file with 2,500,000 chunks.  You'll have to get the sha256 hashes of chunks 0 through 1,999,999, but you'll only have to do the obfuscation and trying for hashes 2,000,000 through 2,500,000.  The method could just return the values for 2,000,000 thru 2,500,000, don't need ALL the addresses for the other sectors.  
*/


addresses := make([]giota.Address, 0, len(sectorDataMap))
for _, sectorDataMapAddress := range sectorDataMap {
addresses = append(addresses, giota.Address(sectorDataMapAddress))
}
request := giota.FindTransactionsRequest{
Addresses: addresses,
}
response, err := api.FindTransactions(&request)

trytesArray, err := api.GetTrytes(response.Hashes)

/* You're on the right track here.  You basically need to verify this:   
1.  For every address in addresses, there is a transaction in trytesArray.Trytes, and there is at least one transaction with that address from within the last epoch (there could be more than 1 with the same address)
*/

//transactionObjects := map[giota.Address][]giota.Transaction{}
transactionInLastEpoch := map[giota.Address][]bool
^You will have to experiment with this

// You could do something like this:  
for _, address := range addresses {
       transactionInLastEpoch[address] = false
}

for _, txObject := range trytesArray.Trytes {
   if txObject.timestamp > timeWeAreComparingAgainst {
      transactionInLastEpoch[txObject.Address] = true
   }
}

/* then iteration through transactionInLastEpoch.  If ALL values are true, the webnode did the PoW.  If any values are false, it lied.  */

```

actions/treasures.go
rfornea: Yeah we'll probably want to do at least like 1000 or so for the launch, but as long as we're saving it in a variable somewhere that's easily changeable, that's what matters.  

@tenisakb For testing purposes, it would be nice if you could write this functionality in such a way that it's easily testable.  I.e. we should be able to verify that if we tell it to verify 5 random transactions, the indexes it will return will all be somewhere in the subsectors.  I.e. the first random transaction is between 0 and 199,000, etc, the second randomly selected value is between 200,000 and 399,999, etc.

rfornea: You could have a separate method that picks these random indexes and just returns them.

actions/treasures.go
rfornea: Do it on the fly.  After mainnet launch we'll explore moving it to a table.

actions/treasures.go
rfornea: @tenisakb @pzhao5 Just so you guys know, all these extra steps just to add a method to the iota_wrappers file is in order for us to be able to mock the method easily.

PR #262, opened 2018-05-04
fix upload_sessions error in batch updates, log every event.

rfornea:
-Segment logs everything.  
-Fixes a small error in actions/upload_sessions.go that was causing the data_maps array to not get populated.  There may be different ways we can approach this, if the way I have fixed it is not suitable.  
https://gophers.slack.com/archives/C3MSAFD40/p1525467579000325

### DIFF
--- a/actions/upload_sessions.go
+++ b/actions/upload_sessions.go
@@ -282,5 +282,5 @@ func (usr *UploadSessionResource) GetPaymentStatus(c buffalo.Context) error {
 }
 
 func sqlWhereForGenesisHashAndChunkIdx(genesisHash string, chunkIdx int) string {
-	return fmt.Sprintf("(genesis_hash = %s AND chunk_idx = %d)", genesisHash, chunkIdx)
+	return fmt.Sprintf("(genesis_hash = '%s' AND chunk_idx = %d)", genesisHash, chunkIdx)
 }
--- a/jobs/claim_unused_prls.go
+++ b/jobs/claim_unused_prls.go
@@ -5,6 +5,7 @@ import (
 	"github.com/oysterprotocol/brokernode/models"
 	"github.com/oysterprotocol/brokernode/services"
 	"github.com/oysterprotocol/brokernode/utils"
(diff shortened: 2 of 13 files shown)


actions/upload_sessions.go
```diff
@@ -282,5 +282,5 @@ func (usr *UploadSessionResource) GetPaymentStatus(c buffalo.Context) error {
 }
 
 func sqlWhereForGenesisHashAndChunkIdx(genesisHash string, chunkIdx int) string {
-	return fmt.Sprintf("(genesis_hash = %s AND chunk_idx = %d)", genesisHash, chunkIdx)
+	return fmt.Sprintf("(genesis_hash = '%s' AND chunk_idx = %d)", genesisHash, chunkIdx)
```
rfornea: This was the sql error i was getting that prompted this fix:
`mysql select many: Error 1054: Unknown column '7c762a2ee5b86dc448dbed91ca75c6cede21622113f9b21e9000b4603455726a' in 'where clause'`.  
Might be some other options:  
https://gophers.slack.com/archives/C3MSAFD40/p1525467579000325

rfornea: When you log the raw query it looks like this:  
```SELECT * from data_maps WHERE (genesis_hash = c08536ccb775919edf7950a209fd5a68cc3a8c555f3209c3a89a0ecb97989ab3 AND chunk_idx = 3) OR (genesis_hash = c08536ccb775919edf7950a209fd5a68cc3a8c555f3209c3a89a0ecb97989ab3 AND chunk_idx = 2) OR (genesis_hash = c08536ccb775919edf7950a209fd5a68cc3a8c555f3209c3a89a0ecb97989ab3 AND chunk_idx = 1) OR (genesis_hash = c08536ccb775919edf7950a209fd5a68cc3a8c555f3209c3a89a0ecb97989ab3 AND chunk_idx = 0)
```

rfornea: cc @pzhao5

pzhao5:
I should add more test for it

PR #266, opened 2018-05-07
Better encryption

rfornea:
-Remove salt
-Remove iv
-Use sha256 hash as the nonce.  Have to trim it according to GCM's nonce size
-Use sha3-256 instead of sha512 hashes as the keys
-Remove deriveKey method and just use the key we tell it to
-Add tests for encryption and decryption methods, we can also use these test vectors if we write encryption/decryption tests on the webnode
-Update CreateTreasurePayload method
-Encrypt and Decrypt methods now return []bytes
-Panic on every single error

### DIFF
--- a/utils/cryptography.go
+++ b/utils/cryptography.go
@@ -3,51 +3,51 @@ package oyster_utils
 import (
 	"crypto/aes"
 	"crypto/cipher"
-	"crypto/rand"
-	"crypto/sha256"
 	"encoding/hex"
 	"hash"
-	"strings"
-
-	"golang.org/x/crypto/pbkdf2"
 )
 
-func deriveKey(passphrase string, salt []byte) ([]byte, []byte) {
-	if salt == nil {
-		salt = make([]byte, 8)
-		// http://www.ietf.org/rfc/rfc2898.txt
-		// Salt.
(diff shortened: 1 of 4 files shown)


utils/cryptography.go
```diff
@@ -3,51 +3,51 @@ package oyster_utils
...
+	keyInBytes, err := hex.DecodeString(key)
+	panicOnErr(err)
+	block, err := aes.NewCipher(keyInBytes)
+	panicOnErr(err)
+	gcm, err := cipher.NewGCM(block)
+	panicOnErr(err)
+	nonceInBytes, err := hex.DecodeString(nonce[0 : 2*gcm.NonceSize()])
+	panicOnErr(err)
+	data := gcm.Seal(nil, nonceInBytes, []byte(secret), nil)
+	return data
```
rfornea: There is a lot of calls to this same panicOnErr function in this file.  Almost every step in the encrypting/decrypting process has the opportunity for errors and I didn't want to ignore them.  I am open to doing this differently if we want.

automyr: This seems fine to me, at least until we've done some more testing with these changes to ensure it's stable.

PR #275, opened 2018-05-09
merge verification, claim

tenisakb:


actions/treasures.go
```diff
@@ -23,65 +21,32 @@ type treasureReq struct {
...
+		verify = false
 	}
 
-	transactionsMap, err := services.FindTransactions(iotaAddr)
-
-	if err != nil || len(transactionsMap) != len(iotaAddr) {
-		// indicate that PoW failure.
+	if verify == true {
+		//msg := EthWrapper.OysterCallMsg{From: "from", To: "to", Amount: 10, PrivateKey:, Gas: 1, TotalWei: 1, Data: [01, 11, 10]}
+		//result := EthWrapper.BuryPrl(msg)
```
rfornea: You'd be calling something like ClaimPRL here, not BuryPRL.  But I'm going to go ahead and approve.

rfornea: @pzhao5 @tenisakb ^

tenisakb: I know we are doing claim but in the past messages was bury.. but understand, will repair

PR #277, opened 2018-05-09
Clean up the verifyTreasure logic

pzhao5:


services/iota_wrappers.go
```diff
@@ -478,11 +479,11 @@ func verifyTreasure(addr []string) (verify bool, err error) {
...
 	if err != nil {
 		raven.CaptureError(err, nil)
-		return verify, err
+		return false, err
 	}
 
 	if len(transactionsMap) != len(iotaAddr) {
-		// indicate that PoW failure.
+		return false, nil
 	}
```
rfornea: We could also add return false, nil on line 494
		if _, hasKey := transactionsMap[iotaAddress]; !hasKey {
			return false, nil
		}

pzhao5: Good catch

PR #283, opened 2018-05-11
Pad treasure chunk, update .env.example, update unit tests

rfornea:
-pad treasure chunk with random trytes characters
-update unit tests
-update .env.example


.env.example
```diff
@@ -40,5 +40,5 @@ MODE="PROD_MODE"
 # May not need these
 
 TEST_MODE_WALLET_ADDRESS="this_can_be_anything"
-TEST_MODE_WALLET_KEY="no_money_no_problems"
```
AaronVasquez: I liked this tag :(

rfornea: Me too.  :(  But in order to be able to handle it the same as the private key payloads, it had to be 64 characters and just numbers, like the private keys.

PR #290, opened 2018-05-13
fix db errors, better logging

rfornea:
-Addresses database errors
-Improved logging
-Adds time-tracking for certain functionality in actions/upload_sessions


actions/upload_sessions.go
```diff
@@ -111,7 +122,7 @@ func (usr *UploadSessionResource) Create(c buffalo.Context) error {
 		// Should we be hardcoding the port?
 		betaURL := req.BetaIP + ":3000/api/v2/upload-sessions/beta"
 		betaRes, err := http.Post(betaURL, "application/json", reqBetaBody)
-
+		defer betaRes.Body.Close() // we need to close the connection
```
rfornea: Adding this didn't fix any errors, but it didn't cause any either.

actions/upload_sessions.go
```diff
@@ -175,11 +204,15 @@ func (usr *UploadSessionResource) Update(c buffalo.Context) error {
...
+			maxChunkIdx = math.Max(maxChunkIdx, float64(chunkIdx))
 		}
 
 		var dms []models.DataMap
-		rawQuery := fmt.Sprintf("SELECT * from data_maps WHERE %s", strings.Join(sqlWhereClosures, " OR "))
-		err := models.DB.RawQuery(rawQuery).All(&dms)
+		//rawQuery := fmt.Sprintf("SELECT * from data_maps WHERE %s", strings.Join(sqlWhereClosures, " OR "))
+		err := models.DB.RawQuery(
+			"SELECT * from data_maps WHERE genesis_hash = ? AND chunk_idx >= ? AND chunk_idx <= ?",
+			uploadSession.GenesisHash, minChunkIdx, maxChunkIdx).All(&dms)
```
rfornea: This may not be the best way to handle this.  But this line is where we were getting the "busy buffer" error before.  Making the query smaller seems to have fixed it.  

As long as we continue to send the chunks in contiguous batches and don't send them out of order or something, this should work.

actions/upload_sessions.go
```diff
@@ -224,9 +257,15 @@ func (usr *UploadSessionResource) Update(c buffalo.Context) error {
...
-			rawQuery = fmt.Sprintf("INSERT INTO data_maps (%s) VALUES %s ON DUPLICATE KEY UPDATE message = VALUES(message), status = VALUES(status), updated_at = VALUES(updated_at)",
+
+			rawQuery := fmt.Sprintf("INSERT INTO data_maps (%s) VALUES %s ON DUPLICATE KEY UPDATE message = VALUES(message), status = VALUES(status), updated_at = VALUES(updated_at)",
 				dbOperation.GetColumns(), strings.Join(sectionUpdatedDms, ","))
+
 			err = models.DB.RawQuery(rawQuery).All(&[]models.DataMap{})
+			for err != nil {
+				time.Sleep(300 * time.Millisecond)
+				err = models.DB.RawQuery(rawQuery).All(&[]models.DataMap{})
+			}
```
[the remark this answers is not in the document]
rfornea: We can also experiment with different batch sizes and see if we can get these errors less often.

PR #291, opened 2018-05-14
Claim unused prls

Astor:
Updated ethereum integration needs to merged to unblock further PRs


AaronVasquez:
Looks like travis is failing because it doesn't compile

PR #292, opened 2018-05-15
Claim prl

Astor:
Added the updated claim prl api and supporting test.

### DIFF
--- a/services/eth_gateway.go
+++ b/services/eth_gateway.go
@@ -30,6 +30,7 @@ type Eth struct {
 	ClaimUnusedPRLs     ClaimPRLs
 	GenerateEthAddr     GenerateEthAddr
 	BuryPrl             BuryPrl
+	ClaimPrl            ClaimPRL
 	SendETH             SendETH
 	SendPRL             SendPRL
 	GetGasPrice         GetGasPrice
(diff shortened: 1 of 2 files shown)


services/eth_gateway.go
```diff
@@ -30,6 +30,7 @@ type Eth struct {
 	ClaimUnusedPRLs     ClaimPRLs
 	GenerateEthAddr     GenerateEthAddr
 	BuryPrl             BuryPrl
+	ClaimPrl            ClaimPRL
```
rfornea: So just so I'm understanding things, ClaimPRL is what will actually call the smart contract claim() method, and ClaimPRLs will be retrieving the unused PRLs from the transaction address?  

We might should change the name of ClaimPRLs back to ClaimUnusedPRLs or call it something else like RetrieveUnusedPRLs, just so it's really clear that it is not invoking the smart contract's claim() function, but just doing a transaction with PRL tokens.

services/eth_gateway.go
```diff
@@ -334,51 +333,34 @@ func buryPrl(msg OysterCallMsg) (bool) {
...
-		}
+	if err != nil {
+		return false
 	}
+	// pull signed transaction
+	ts := types.Transactions{signedTx}
 
-	return nil
+	// successful contract message call
+	if ts.Len() > 0 {
```
pzhao5: nit return ts.Len() > 0

PR #298, opened 2018-05-18
Add NotifyBeta API endpoint for Alpha to call

pzhao5:
Allow Alpha to call NotifyBeta once the ETH transfer is completed and update Beta's payment status.

This API should only for Alpha Node.
Next PR: Add logic on the alpha side to wait for Transfer and call NotifyBeta API endpoint

### DIFF
--- a/actions/app.go
+++ b/actions/app.go
@@ -71,6 +71,7 @@ func App() *buffalo.App {
 		apiV2.POST("upload-sessions", uploadSessionResource.Create)
 		apiV2.PUT("upload-sessions/{id}", uploadSessionResource.Update)
 		apiV2.POST("upload-sessions/beta", uploadSessionResource.CreateBeta)
+		apiV2.POST("upload-sessions/notifyBeta", uploadSessionResource.NotifyBeta)
 		apiV2.GET("upload-sessions/{id}", uploadSessionResource.GetPaymentStatus)
 
 		// Webnodes
--- a/actions/upload_sessions.go
+++ b/actions/upload_sessions.go
@@ -15,6 +15,7 @@ import (
 	"github.com/pkg/errors"
 	"math"
 	"net/http"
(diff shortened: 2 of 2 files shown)


AaronVasquez:
The beta broker should attach a transfer listener itself so it doesn't rely on alpha to notify it. All alpha has to do is send half the PRLs to Beta once it receives it from the client.

pzhao5:
We decide to change from subscribeToTransfer to waitForTransfer (a blocking call). And Alpha will wait for it to complete.

Should Beta also do waitForTransfer?

pzhao5:
Alpha still need an API endpoint to send Beta a POST request

rfornea:
>Alpha still need an API endpoint to send Beta a POST request   

I may be misunderstanding.  If both Alpha and Beta separately do waitForTransfer, then what would Alpha be making a POST request to Beta for?  

@AaronVasquez  This just occurred to me, but it might be a mistake to have alpha and beta waitForTransfer separately.  This would mean that Beta would not start processing the file until Alpha has sent it the money.  Which means alpha would have several minutes of headstart on the file.

rfornea:
If we still want them to wait separately, we might want Beta to simply wait for the Alpha --> Beta transaction to be initiated, rather than waiting for it to actually arrive/confirm.  In this way, Beta would be starting at around the same time as Alpha.

rfornea:
Beta would have to rely on Alpha using enough gas to ensure the transaction will get through.  But long-term, Beta could report Alpha to the DRS if it didn't.

AaronVasquez:
If we want Beta to start at the same time, it can listen for client -> alpha confirmed (same event as alpha), and assume alpha will be a good actor (DRS can enforce this).

@pzhao5 I misunderstanding the problem. Why does alpha need to notify beta?

PR #300, opened 2018-05-18
Add waitForTransfer method for both Alpha/Beta

pzhao5:


actions/upload_sessions.go
```diff
@@ -358,3 +361,24 @@ func (usr *UploadSessionResource) GetPaymentStatus(c buffalo.Context) error {
...
+	} else {
+		session.PaymentStatus = models.PaymentStatusError
+	}
+
+	if err := models.DB.Save(&session); err != nil {
+		return
+	}
+
+	// TODO(pzhao5): NotifyBeta on payment status
+}
```
rfornea: I tend to agree with what Aaron mentioned in the other PR.  This method could just be called waitForTransfer and we could call it from both Create and Createbeta.

PR #303, opened 2018-05-18
Better handling error case in EthGateway

pzhao5:
In A few places we did not handle err at all and we return nil object from sharedClient() method.
This will lead to panic due to nil object.

### DIFF
--- a/services/eth_gateway.go
+++ b/services/eth_gateway.go
@@ -117,21 +117,19 @@ func sharedClient(netUrl string) (c *ethclient.Client, err error) {
 	mtx.Lock()
 	defer mtx.Unlock()
 
-	if client != nil {
-		// override to allow custom node url
-		if len(netUrl) > 0 {
-			ethUrl = netUrl
-		}
-		c, err = ethclient.Dial(ethUrl)
-		if err != nil {
-			fmt.Println("Failed to dial in to Ethereum node.")
-			raven.CaptureError(err, nil)
-			return
-		}
-		// Sets Singleton
-		client = c
+	// override to allow custom node url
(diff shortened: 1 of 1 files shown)


services/eth_gateway.go
```diff
@@ -212,17 +210,25 @@ func getCurrentBlock() (*types.Block, error) {
 // WaitForTransfer is blocking call that will observe on brokerAddr on transfer on ETH.
 // If it is completed return true, otherwise, return false (or time-out)
```
rfornea: Instead of returning booleans, would it be an option in some of these places to just do 'return err' so it's obvious to someone watching the broker what the error was?  Something else that would be really helpful is to send all these errors to Sentry so we can track them.  You can to that like this:  
		if err != nil {
			raven.CaptureError(err, nil)
                        return err
		}

pzhao5: I think in some places, it makes sense to return boolean rather than an error. Since caller does not really care about the error and it won't be able to do much. 

But in a case by case, any method you think we should return err instead, please let me know.

For example, sendETH return a string, instead, it should return string, error. I am planning to fix it in the next PR.

AaronVasquez: I agree with Rebel. Returning an err is more clear to the caller that the function can fail. A success boolean isn't clear from just reading the function signature. Checking for err seems to be a Golang idiom too.

rfornea: I'm going to make a task to add Sentry logging throughout the brokernode.  There's quite a few files where we aren't going it, and we probably should be doing it everywhere.  It'll help us a lot in the coming week.

PR #312, opened 2018-05-20
Update how storage cost is calculated

rfornea:
-Properly calculates the PRL invoice
-Adds unit tests

### DIFF
--- a/models/upload_sessions.go
+++ b/models/upload_sessions.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"github.com/oysterprotocol/brokernode/utils"
-	"math"
 	"os"
 	"time"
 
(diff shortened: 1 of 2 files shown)


models/upload_sessions_test.go
```diff
@@ -295,3 +296,94 @@ func (ms *ModelSuite) Test_GetTreasureIndexes() {
...
+	}
+
+	vErr, err := u.StartUploadSession()
+	invoice := u.GetInvoice()
+	ms.Nil(err)
+	ms.Equal(0, len(vErr.Errors))
+
+	// expecting to be charged for 2 full sectors
+	// should be same price as previous test
+	ms.Equal(decimal.NewFromFloatWithExponent(0.09375000000000001, -16), invoice.Cost)
```
rfornea: The 2nd parameter in NewFloatWithExponent is the exponent, and if you give it `decimal.NewFromFloatWithExponent(0.09375, -16)` it ignores the -16 and the exponent will be -5.  And when you do `decimal.NewFromFloatWithExponent(0.0937500000000000, -16)` it truncates back to -5.  Hence the additional 1, which gets cut off but it makes the result have the right exp value.

models/upload_sessions.go
```diff
@@ -68,6 +66,8 @@ const (
 	TreasureBuried
 )
 
+var StoragePeg = 64 // GB per year per PRL; TODO: query smart contract for real storage peg
```
pzhao5: Or you could make it decimal number rather than int

rfornea: Okay, I'll change it.

rfornea:
Making a small update to this in the morning

PR #314, opened 2018-05-21
Split half of the PRL with Beta

pzhao5:


actions/upload_sessions.go
```diff
@@ -367,22 +370,37 @@ func sqlWhereForGenesisHashAndChunkIdx(genesisHash string, chunkIdx int) string
...
+	session.PaymentStatus = paymentStatus
 	if err := models.DB.Save(&session); err != nil {
+		raven.CaptureError(err, nil)
 		return
 	}
-	// TODO(pzhao5): do other stuff here
+
+	// Alpha send half of it to Beta
+	if isAlpha && paymentStatus == models.PaymentStatusConfirmed {
+		var splitedAmount big.Int
```
rfornea: nit:  splittedAmount or just splitAmount.  Otherwise looks good.

actions/upload_sessions.go
```diff
@@ -367,22 +370,37 @@ func sqlWhereForGenesisHashAndChunkIdx(genesisHash string, chunkIdx int) string
...
-
+	session.PaymentStatus = paymentStatus
 	if err := models.DB.Save(&session); err != nil {
+		raven.CaptureError(err, nil)
 		return
 	}
-	// TODO(pzhao5): do other stuff here
+
+	// Alpha send half of it to Beta
+	if isAlpha && paymentStatus == models.PaymentStatusConfirmed {
```
AaronVasquez: Instead of passing in `isAlpha` can we use the session type from the uploadSession struct to determine if this is the alpha or beta broker? It would reduce the responsibility of the caller to know whether to pass true of false.

AaronVasquez:
(Not for this PR) Do the listeners survive a restart if the brokernode goes down for whatever reason?

pzhao5:
To answer your question, Nope. We need to have startup process to handle this.

services/eth_gateway.go
```diff
@@ -234,16 +233,16 @@ func getCurrentBlock() (*types.Block, error) {
 }
 
 // WaitForTransfer is blocking call that will observe on brokerAddr on transfer on ETH.
-// If it is completed return true, otherwise, return false (or time-out)
-func waitForTransfer(brokerAddr common.Address) (bool, error) {
+// If it is completed return number of PRL.
```
AaronVasquez: Number fo PRLs should be a decimal since we can have partial PRLs (eg: 1.5)

pzhao5: Confusing, since part of our code use big.Int, part of our code use decimal. We should just use one

rfornea: Currently we're using decimal for the Cost/Invoice, but we could certainly change that.  This might be a dumb question but do big.Ints let you have a decimal place?  Or do they just show a much smaller unit of the currency?  

If we have to send 1.5 PRL I'd rather it show up as 1.5 PRL for readability, would be my personal preference.

AaronVasquez: I was thinking that since we know PRLs have 18 decimal precision, we would store this as an int, then divide by 10^18 we want to convert PRLs. It complicates things a bit.

If Golang has arbitrary precision decimals (or at least to 18 decimals), I would prefer to use that so we don't get confused with the conversion.

I'm going to approve this PR, and we can convert everything to something consistent in a future PR.

AaronVasquez: Looks like we can use this: https://golang.org/pkg/math/big/#NewFloat

pzhao5: https://github.com/shopspring/decimal#why-dont-you-just-use-bigrat
Probably to stick with decimal

AaronVasquez: SGTM

PR #316, opened 2018-05-21
Encrypt Eth Keys instead of storing naked keys

rfornea:
-encrypts the eth keys so we aren't just storing them naked in the dbs.  
-for keys and nonces, I'm using hashes of the session ID and hashes of the CreatedAt time for sessions (if encrypting an ethKey for a session) or for chunks (if encrypting eth keys for treasures)
-buffalo suggests using c.Error instead of c.Render for errors

FYI, this PR means that we'll have to decrypt the eth key of the session to be able to use it for transactions.  But there's a method for that.

### DIFF
--- a/actions/upload_sessions.go
+++ b/actions/upload_sessions.go
@@ -116,7 +116,7 @@ func (usr *UploadSessionResource) Create(c buffalo.Context) error {
 		if err != nil {
 			fmt.Println(err)
 			raven.CaptureError(err, nil)
-			c.Render(400, r.JSON(map[string]string{"Error starting Beta": err.Error()}))
+			c.Error(400, err)
 			return err
 		}
 
@@ -129,7 +129,7 @@ func (usr *UploadSessionResource) Create(c buffalo.Context) error {
 		if err != nil {
 			fmt.Println(err)
 			raven.CaptureError(err, nil)
-			c.Render(400, r.JSON(map[string]string{"Error starting Beta": err.Error()}))
+			c.Error(400, err)
 			return err
 		}
 		betaSessionRes := &uploadSessionCreateBetaRes{}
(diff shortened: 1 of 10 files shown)


actions/upload_sessions.go
```diff
@@ -183,12 +199,18 @@ func (usr *UploadSessionResource) Update(c buffalo.Context) error {
...
+	if err != nil {
 		fmt.Println(err)
 		raven.CaptureError(err, nil)
-		c.Render(400, r.JSON(map[string]string{"Error finding session": errors.WithStack(err).Error()}))
+		c.Error(400, err)
 		return err
 	}
+	if uploadSession == nil {
+		err := errors.New("Error finding sessions")
+		c.Error(400, err)
```
pzhao5: also need to captureError

actions/upload_sessions.go
```diff
@@ -350,8 +387,13 @@ func (usr *UploadSessionResource) GetPaymentStatus(c buffalo.Context) error {
...
 
-	if (err != nil || session == models.UploadSession{}) {
-		//TODO: Return better error response when ID does not exist
+	if err != nil {
+		c.Error(400, err)
+		raven.CaptureError(err, nil)
+		return err
+	}
+	if (session == models.UploadSession{}) {
+		c.Error(400, errors.New("Did not find session that matched id"))
```
pzhao5: capture the error

services/eth_gateway.go
```diff
@@ -143,6 +147,21 @@ func generateEthAddr() (addr common.Address, privateKey string, err error) {
...
 
+// Generate an array of eth keys
+func generateKeys(numKeys int) []string {
+	var keys []string
+	for i := 0; i < numKeys; i++ {
+		key := ""
+		if oyster_utils.BrokerMode == oyster_utils.TestModeDummyTreasure {
+			key = os.Getenv("TEST_MODE_WALLET_KEY")
+		} else {
+			_, key, _ = generateEthAddr()
```
pzhao5: Do you need to check error, if there is something wrong, you should not add the key into keys

pzhao5:
Otherwise, look good

services/eth_gateway.go
rfornea: I'll add it

actions/upload_sessions.go
rfornea: I'll add the captures

actions/upload_sessions.go
```diff
@@ -155,7 +155,13 @@ func (usr *UploadSessionResource) Create(c buffalo.Context) error {
 		fmt.Println(err)
 	}
 
-	privateKeys := services.EthWrapper.GenerateKeys(len(mergedIndexes))
+	privateKeys, err := services.EthWrapper.GenerateKeys(len(mergedIndexes))
+	if err != nil {
+		err := errors.New("Could not generate eth keys: " + err.Error())
+		fmt.Println(err)
```
rfornea: For this error and the other new error added to this file, I don't send the error to sentry because it gets captured in the generateKeys method.

PR #330, opened 2018-05-23
Centralize ENV var accessing

AaronVasquez:


rfornea:
I approved before I thought about this.  We might want to throw an error or log something if there is nothing for host_ip.

utils/env.go
```diff
@@ -0,0 +1,55 @@
...
+	if err == nil {
+		for _, netInterfaceAddress := range netInterfaceAddresses {
+			networkIp, ok := netInterfaceAddress.(*net.IPNet)
+			if ok && !networkIp.IP.IsLoopback() && networkIp.IP.To4() != nil {
+				ip := networkIp.IP.String()
+				fmt.Println("Resolved Host IP: " + ip)
+				return ip
+			}
+		}
+	}
```
rfornea: My only concern here would be, is this logic going to get the ip address of the actual host machine or is it going to be the ip address of the docker container?

AaronVasquez: Yes, I forgot about that....I'm going to put a pause on this PR until we can think of a better solution. Otherwise we can just stick with a manual solution for now.

PR #332, opened 2018-05-24
Add Util function to convert Wei to Prl and reverse

pzhao5:


utils/utils.go
```diff
@@ -210,3 +214,16 @@ func RandSeq(length int, sequence []rune) string {
...
+
+/* Convert PRL unit to wei unit. */
+func ConvertToWeiUnit(prl *big.Float) *big.Int {
+	f := new(big.Float).Mul(prl, big.NewFloat(float64(PrlInWeiUnit)))
+	wei, _ := f.Int(new(big.Int)) // ignore the accuracy
+	return wei
+}
+
+/* Convert wei unit to PRL unit */
+func ConvertToPrlUnit(wei *big.Int) *big.Float {
```
AaronVasquez: Nit: Rename to `ConverFromWeiUnit` so we are consistent with the `web3` API

PR #333, opened 2018-05-24
Check balance first before wait for Transfer

pzhao5:
If balance is there, then don't need to wait further for the transfer

### DIFF
--- a/services/eth_gateway.go
+++ b/services/eth_gateway.go
@@ -229,7 +229,7 @@ func checkBalance(addr common.Address) *big.Int {
 	// connect ethereum client
 	client, err := sharedClient("")
 	if err != nil {
-		log.Fatal("Could not initialize shared client")
+		return big.NewInt(0)
 	}
 
 	balance, err := client.BalanceAt(context.Background(), addr, nil) //Call(&bal, "eth_getBalance", addr, "latest")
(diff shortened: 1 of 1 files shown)


services/eth_gateway.go
```diff
@@ -266,6 +266,12 @@ func getCurrentBlock() (*types.Block, error) {
 // WaitForTransfer is blocking call that will observe on brokerAddr on transfer on ETH.
 // If it is completed return number of PRL.
 func waitForTransfer(brokerAddr common.Address) (*big.Int, error) {
+	balance := checkBalance(brokerAddr)
+	if balance.Int64() > 0 {
+		// Has balance already, don't need to wait for it.
+		return balance, nil
+	}
```
rfornea: Does checkBalance return the balance of PRL or the balance of ETH?  Or both?

pzhao5: Ideally, it should be PRL, and we use it in the codebase. Maybe Astor could have a better answer

AaronVasquez: @Astor ^

rfornea: Ultimately we will need methods to get balances of both ETH and PRL. I created a Zenhub issue to have methods for both.   

If everyone is cool with it we could just go ahead and approve this, and splitting out those methods could be something we do soonish. We will have to look through the codebase at where the current method is being used and make sure it is replaced with the correct one.

Astor: It checks the for both balances.

PR #343, opened 2018-05-25
Update brokernode to work with webinterface's updated encryption

rfornea:
-updates the brokernode to work with webinterface's updated encryption
-uses valid hex strings wherever we are using a test genesis hash (we should break this value out in some future review)

Merge at the same time as corresponding webinterface review, as this is a breaking change.


rfornea:
If anything is wrong with this PR feel free to pull down the branch and update it.

PR #353, opened 2018-05-28
CheckPRLBalance returns in Wei unit. So not need to convert it again

pzhao5:


rfornea:
Oops, didn't realize the tests had failed.

PR #357, opened 2018-05-29
Disables POW with a flag

AaronVasquez:


AaronVasquez:
I left it out to make the default (nil value) to not disable POW

PR #367, opened 2018-05-30
Remove dup raven.Capture

pzhao5:
It has been logged in the caller method. So don't need to log it again.


AaronVasquez:
Not merging yet since we have a freeze while we deploy

PR #368, opened 2018-05-30
Remove .env.example file

pzhao5:


.env.example
```diff
@@ -27,7 +27,7 @@ HOST_IP=""
 # Without these you cannot receive PRL revenue
 
 MAIN_WALLET_ADDRESS="(your broker’s main wallet address)"
-MAIN_WALLET_KEY="(privatekey to broker’s main wallet"
+MAIN_WALLET_KEY="bc07ec20ceedff112f1498a63f6da115a78d6b26fb6ec282bf8b3f45e3358fdf"
```
rfornea: I think that this file can actually be deleted.  I believe we renamed the file .env.test and have not deleted this .env.example file yet.

AaronVasquez:
Btw, the `.env` file isn't committed. Idk if you were trying to commit changes there.

PR #372, opened 2018-05-30
Key update

Astor:
fixes for the keystore and env parsing and nonce test.


services/eth_gateway.go
```diff
@@ -353,14 +356,34 @@ func isPending(txHash common.Hash) bool {
...
 	client, _ := sharedClient()
 	// web3.eth.blockNumber-web3.eth.getTransaction("0xa92c69ebc71f46ef73a7bcf4b7a17aaa1b8daa7f6036e3c3374b62589ac1b8d3").blockNumber
+	//block, _ := getCurrentBlock()
+	//blockNumber := block.Number()
+	
 	// get transaction
-	txBlockNumber, err := client.TransactionCount(context.Background(), txHash)
+	tx, isPending, err := client.TransactionByHash(context.Background(), txHash)
 	if err != nil {
 		fmt.Printf("Could not get transaction by hash")
```
rfornea: We should send this error to sentry.  raven.CaptureError(err, nil)

services/eth_gateway.go
```diff
@@ -353,14 +356,34 @@ func isPending(txHash common.Hash) bool {
...
-	fmt.Printf("tx block number : %v\n", txBlockNumber)
-	return big.NewInt(0).SetUint64(uint64(txBlockNumber)), nil
+	if isPending {
+		fmt.Println("transaction is pending")
+	}
+	fmt.Printf("tx block number : %v\n", tx.Nonce())
+	data := tx.Data()
+	tx.UnmarshalJSON(data)
+	if err != nil {
+		fmt.Printf("could not unmarshal transaction data")
```
rfornea: Same here

services/eth_gateway.go
rfornea: Or better yet just use the oyster_utils.LogIfError(err, nil) method.

PR #377, opened 2018-05-31
Adding additional tag for Raven

pzhao5:


utils/utils.go
```diff
@@ -38,6 +39,25 @@ const (
...
+	}
+	logErrorTags = map[string]string{
+		"mode":       os.Getenv("MODE"),
+		"host_ip":    os.Getenv("HOST_IP"),
+		"osyter_pay": isOysterPay,
+	}
+}
+
+/*SetLogInfoForDatabaseUrl updates db_url for log info.*/
+func SetLogInfoForDatabaseUrl(dbUrl string) {
```
rfornea: My only concern with this PR is what this value is going to be when this code is being run on a dev broker.  In that situation the db would be local in a container on the broker, and the person running the broker most likely would not have specified a value for it in their .env file.  For those cases I'm guessing DB.URL() would just be nil or something?  We could do something like:
 if os.Getenv("DATABASE_URL") == "" { db_url = "local" } else { db_url = os.Getenv("DATABASE_URL") }  

Though I suppose if we see it as nil in Sentry, we could just assume that the error is coming from a dev broker.

pzhao5: for dev build: this will be: 
root:secret@(db_test:3306)/brokernode_test?parseTime=true&multiStatements=true&readTimeout=1s

rfornea: Okay, cool

PR #379, opened 2018-06-01
prometheus implement

tenisakb:


Dockerfile
```diff
@@ -44,6 +44,9 @@ WORKDIR $GOPATH/src/github.com/oysterprotocol/brokernode
 
 RUN go get -u github.com/gobuffalo/buffalo/buffalo
 
+# # Prometheus client:
+RUN go get -d -v github.com/prometheus/client_golang/prometheus
```
AaronVasquez: Isn't this fetched from ` RUN go get -t -d -v ./...` below?

actions/app.go
```diff
@@ -24,6 +25,18 @@ var app *buffalo.App
 // Visible for Unit Test
 var IotaWrapper = services.IotaWrapper
 var EthWrapper = services.EthWrapper
+var PrometheusWrapper = services.PrometheusWrapper
```
AaronVasquez: Can we move this and the lines below to `services/prometheus.go`?

actions/transaction_genesis_hashes.go
```diff
@@ -46,6 +47,12 @@ type transactionGenesisHashUpdateRes struct {
 
 // Creates a transaction.
 func (usr *TransactionGenesisHashResource) Create(c buffalo.Context) error {
+	start := time.Now()
+
+	defer func() {
```
AaronVasquez: Does this need to be wrapped in a closure?

Dockerfile
tenisakb: I dont know what you mean, its needy to have it here to run prometheus.. how would you like to have it?

actions/transaction_genesis_hashes.go
tenisakb: Would you like to have here only function from service? In this function will be included histogram? But I dont know then how to work with defer function.

Dockerfile
tenisakb: I will take a look what lines below do exactly nad will try to remove

PR #381, opened 2018-06-03
Binary encoded strings instead of trytes (merge/deploy with corresponding webinterface review)

rfornea:
-broker can now receive the message as trytes or as a binary string
-instead of casting using giota.Trytes() or giota.Address(), the giota library recommends we use giota.ToAddress() and giota.ToTrytes() which has built-in validity checking
-adds version to upload_sessions and completed uploads

I was going to add it to the data_maps table but thought of some possible issues which I mentioned in the mainnet channel.  

Also was going to add a separate migrations file for this change, something like 20180603543234_add_version_to_sessions.up.fizz or something, but I got complaints when I tried to run migrations this way.  If we can't get away with modifying the existing migration files, I will explore this further.  Let's discuss this before we deploy.

### DIFF
--- a/utils/trytes_conversion.go
+++ b/utils/trytes_conversion.go
@@ -12,6 +12,8 @@ var (
 	TrytesAlphabet = []rune("9ABCDEFGHIJKLMNOPQRSTUVWXYZ")
 )
 
+const StopperTryte = "A"
+
 func AsciiToTrytes(asciiString string) (string, error) {
 	var b strings.Builder
 
@@ -83,6 +85,21 @@ func TrytesToBytes(t giota.Trytes) []byte {
 		output = append(output, c)
 	}
 	return output
+}
+
+func ChunkMessageToTrytesWithStopper(binString string) giota.Trytes {
+	trytes := RunesToTrytes([]rune(binString))
+	return giota.Trytes(trytes + StopperTryte)
+}
+
+func RunesToTrytes(r []rune) giota.Trytes {
+	var output string
+	for _, c := range r {
+		v1 := c % 27
+		v2 := (c - v1) / 27
+		output += string(TrytesAlphabet[v1]) + string(TrytesAlphabet[v2])
+	}
+	return giota.Trytes(output)
 }
 
 func BytesToTrytes(b []byte) giota.Trytes {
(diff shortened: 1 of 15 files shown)


utils/trytes_conversion.go
```diff
@@ -85,6 +87,21 @@ func TrytesToBytes(t giota.Trytes) []byte {
 	return output
 }
 
+func ChunkMessageToTrytesWithStopper(binString string) giota.Trytes {
+	trytes := RunesToTrytes([]rune(binString))
+	return giota.Trytes(trytes + StopperTryte)
+}
+
+func RunesToTrytes(r []rune) giota.Trytes {
```
automyr: I know you had some issues when trying to preserve the data when doing straight bytes -> trytes without the runes, so perhaps you already took care of this somewhere else in the code, but runes seem to be an alias for the `int32` type to allow programmers to unambiguously know which codepoint (as referenced by utf-8, such as U+0031) you're referring to, and it seems like 1 rune -> 1 tryte (? correct me if I'm reading this wrong). Each tryte has only 3 bits of info, and each rune has 32 bits - so it seems really wasteful to do this unless it's a temporary change than only exists in memory. ie: webinterface sends bytes -> broker stores bytes -> bytes are transformed first to runes and then to trytes -> trytes are uploaded to the tangle.

Perhaps I'm making too many assumptions here though

rfornea: There will be another PR on this same topic in a day or so, to actually store the bytes in the DB and then just convert them to trytes on the fly when we need them.  At the moment, I'm still storing the actual trytes in the DB.  However it's not taking up anymore space than our current implementation, since we're already storing the trytes.  

I want to handle storing the bytes and then converting to trytes on the fly in a separate PR, since there's a lot of places where we use the Message field, and I want to test it pretty thoroughly.

rfornea: To clarify this PR is doing this:
webinterface sends bytes -> broker converts to trytes and stores as trytes -> trytes are uploaded to the tangle.

The next PR will be doing this:
webinterface sends bytes -> broker stores bytes -> bytes are transformed first to runes and then to trytes -> trytes are uploaded to the tangle.

automyr:
Well, I think we can merge it now 👍

PR #394, opened 2018-06-10
Disable JobRunner during unit test

pzhao5:


jobs/job_runner.go
```diff
@@ -24,8 +25,11 @@ var (
...
 
 func init() {
-	registerHandlers(OysterWorker)
+	if enabled, err := strconv.ParseBool(os.Getenv("JOB_RUNNER")); err == nil && !enabled {
+		return
+	}
 
+	registerHandlers(OysterWorker)
 	doWork(OysterWorker)
 }
```
rfornea: I am not sure if this is going to do what we need.  What we ultimately need is a way to set this to false as soon as we start the unit tests, and then set it back to true as soon as the tests finish.    

I am not sure if there is a good way to have something run before all the tests, and have something run after all the tests.

rfornea: I'm going to go ahead and approve.  This will be useful for Travis, if nothing else.

pzhao5: The unit test and prod should have a different env. 

For prod, we don't need to set this field. For unit test, we just disable this.

PR #397, opened 2018-06-10
Disable raven log in Unit test

pzhao5:


utils/utils.go
```diff
@@ -39,9 +39,21 @@ const (
...
+// Enable/Disalbe raven reporting.
+var isRavenEnabled bool = true
 var logErrorTags map[string]string
 
 func init() {
+	// Check whether current is in unit test mode.
+	// If provided -v, then it would be -test.v=true. By default, it would output to a log dir
+	// with the following format: -test.testlogfile=/tmp/go-build797632719/b292/testlog.txt
+	for _, v := range os.Args {
+		if strings.HasPrefix(v, "-test.") {
```
rfornea: should we add `|| strings.HasPrefix(v, "_test.")` ?  

Is it looking for the end of the filename or package name?

pzhao5: probably not. since "-test." is a valid args while "_test." is not.

PR #408, opened 2018-06-11
[WIP] Deprecate msg

AaronVasquez:
WARNING: Don't merge this yet. Still a WIP.

### DIFF
--- a/models/data_maps.go
+++ b/models/data_maps.go
@@ -5,20 +5,19 @@ import (
 	"crypto/sha512"
 	"encoding/hex"
 	"encoding/json"
-	"github.com/getsentry/raven-go"
-	"golang.org/x/crypto/sha3"
+	"fmt"
 	"math/rand"
-
+	"strconv"
+	"strings"
 	"time"
 
-	"github.com/oysterprotocol/brokernode/utils"
-
-	"fmt"
+	"github.com/getsentry/raven-go"
 	"github.com/gobuffalo/pop"
 	"github.com/gobuffalo/uuid"
 	"github.com/gobuffalo/validate"
 	"github.com/gobuffalo/validate/validators"
-	"strings"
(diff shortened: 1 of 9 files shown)


models/data_maps.go
```diff
@@ -150,6 +160,11 @@ func (d *DataMap) DecryptEthKey(encryptedKey string) (string, error) {
 	return hex.EncodeToString(decryptedKey), nil
 }
 
+func (d *DataMap) generateMsgId() string {
```
pzhao5: I don't like this because it require you to save DataMap first before getting this key. This requires SQL operation. For example, if you need to get a particular data, via Genesis and chunkID, it is very fast to compute it. But now, you had to do a look up.

services/kv_store.go
```diff
@@ -112,7 +111,20 @@ func BatchSet(kvs *KVPairs) error {
 	return err
 }
 
-/*GenKvStoreKey returns the key for inserting to KV-Store for data_maps.*/
-func GenKvStoreKey(gensisHash string, chunkIdx int) string {
-	return fmt.Sprintf("%s_%d", gensisHash, chunkIdx)
+func BatchDelete(ks *KVKeys) (err error) {
```
pzhao5: can you revert this. Since I will have mine ready

AaronVasquez: yeah, didn't notice you already did this

pzhao5: Also this will break if too many pending transaction

models/data_maps.go
AaronVasquez: This is to ensure that a dataMap will always have a `MsgID` before being saved. This is safer than always generating the key incase the way to generate a key is changed. Also the key for a datamap should be immutable once it's created, this is better documented as a field in the DB.

AFAIK, you will always have a datamap before using the binary data. I don't think we access the binary data outside of the context of a datamap.

PR #410, opened 2018-06-12
Handle Txn too big error

pzhao5:


services/kv_store.go
```diff
@@ -123,14 +143,35 @@ func BatchDelete(ks *KVKeys) error {
...
+				txn = badgerDB.NewTransaction(true)
+				e = txn.Delete([]byte(key))
 			}
 		}
-		return nil
-	})
+
+		if e != nil {
+			err = e
+			break
```
AaronVasquez: Should we continue and try to set as many as possible?

PR #411, opened 2018-06-12
Just disable Job run in test

pzhao5:
It does not make sense to have job run in the background during unit test. For unit test for job run, they should manually enable the each test.

This will reduce some sentry.io error

### DIFF
--- a/jobs/job_runner.go
+++ b/jobs/job_runner.go
@@ -4,8 +4,8 @@ import (
 	"os"
 	"reflect"
 	"runtime"
-	"time"
 	"strconv"
+	"time"
 
 	"github.com/gobuffalo/buffalo/worker"
 	"github.com/oysterprotocol/brokernode/services"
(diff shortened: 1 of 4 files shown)


rfornea:
I don't get how this is going to work.  I think setting JOB_RUNNER=false in .env.test is not going to do anything.  The .env.test file is just there to be an example for people to make their .env files, and for travis to copy over when those tests get run.  I don't think the Golang code actually uses the .env.test file, even when the unit tests are running.  Are people supposed to put this in their .env file?  To be able to use a broker normally for uploads and downloads, you'd have to have the job runner enabled, but if a person wanted to run the unit tests, would they have to disable and then re-enable the setting everytime?

jobs/job_runner.go
```diff
@@ -20,13 +20,15 @@ const (
...
-	EthWrapper  = services.EthWrapper
-	PrometheusWrapper  = services.PrometheusWrapper
+	IotaWrapper       = services.IotaWrapper
+	EthWrapper        = services.EthWrapper
+	PrometheusWrapper = services.PrometheusWrapper
 )
 
 func init() {
-	if enabled, err := strconv.ParseBool(os.Getenv("JOB_RUNNER")); err == nil && !enabled {
+	enabled, err := strconv.ParseBool(os.Getenv("JOB_RUNNER"))
```
AaronVasquez: Since you already have `IsInUnitTest`, can we just use that? I don't think we'd ever disable a runner outside of tests.

pzhao5: Sure.

pzhao5:
@rfornea, you should never enable job runner inside the Unit test. Because when your test is completed, the job could still run and the data base has already been delete.

Providing that flag is to allow pusher to manually disable job-runner if needed. I won't expect everyone to put JOB_RUNNER=false in .env file but at least for Travis, it would be nice to disable. I want to make it harder to enable it during unit test since we get a lot of failure from these background job in Unit test mode.

rfornea:
Okay, I see what you are talking about.

AaronVasquez:
looks like you have a merge conflict

PR #413, opened 2018-06-12
Dev docker

AaronVasquez:


README.md
```diff
@@ -35,6 +38,7 @@ brokernode# buffalo test
 ```
```
rfornea: Do we need some sort of database.dev.yml file as well?

AaronVasquez: This worked for me. The yml file in the `dev-dbs` branch looks to be the same as what's on master.

AaronVasquez:
Actually, don't approve yet, it started failing locally :(

README.md
rfornea: They are some differences.  Configuration for test is the same, but development and production are different.

AaronVasquez: still a little weird how the tests pass on travis, but not locally. Ill try copy and pasting the database.yml and see if that fixes it for me

PR #419, opened 2018-06-16
Final round for BadgerDB

pzhao5:


pzhao5:
Look like the _test table is not found.

PR #434, opened 2018-06-23
Add msgStatus column in DataMaps

pzhao5:
msgStatus will be represent the status of message. Does brokernode hve it or not. It will enable brokernode to do efficient query such as give me all the dataMaps object where its message is uploaded.

Follow up PR will use this field.


models/data_maps.go
```diff
@@ -50,6 +50,7 @@ type DataMap struct {
 	NodeType       string    `json:"nodeType" db:"node_type"`
 	Message        string    `json:"message" db:"message"`
 	MsgID          string    `json:"msgId" db:"msg_id"`
+	MsgStatus      int       `json:"msgStatus" db:"msg_status"`
```
rfornea: In the next PR we could make this an enum like the Status field in this same model.

PR #435, opened 2018-06-24
Adding retry logic to try it best to avoid Deadlock

pzhao5:
See here for deadlock:
https://stackoverflow.com/questions/2332768/how-to-avoid-mysql-deadlock-found-when-trying-to-get-lock-try-restarting-trans

Solution:
1) Adding retry logic
2) Use select method to create a temp table and then update based on that temp table only.


rfornea:
Everything looks fine here. I'm going to restart the build/s a couple of times to see if they'll both turn green.

PR #445, opened 2018-06-25
Treasure status column addition should be in its own file

rfornea:
This column addition should have been in its own file


migrations/20180625135525_add_treasure_status_to_stored_genesis_hashes.up.fizz
```diff
@@ -0,0 +1 @@
+add_column("stored_genesis_hashes", "treasure_status", "integer", {})
```
AaronVasquez: Wouldn't this need it's own `down` migration file too?

rfornea: I think you're right.  I can add that in my next PR.

PR #447, opened 2018-06-26
Add MsgStatus

pzhao5:
Use MsgStatus as way to improve performance of SQL query. Otherwise, we need to cross query.

TODO: assign Uploaded value in the upcoming PR

### DIFF
--- a/models/data_maps.go
+++ b/models/data_maps.go
@@ -39,8 +39,11 @@ const (
 )
 
 const (
+	// Default value before adding msg_status column.
 	MsgStatusUnmigrated int = iota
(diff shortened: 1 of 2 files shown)


models/data_maps.go
```diff
@@ -38,6 +38,15 @@ const (
 	Error = -1
 )
 
+const (
+	// Default value before adding msg_status column.
+	MsgStatusUnmigrated int = iota
```
rfornea: We usually start these with `iota + 1` in most of the rest of the model files, since 0 is the default for an int column when nothing is assigned.

PR #451, opened 2018-06-27
Move CompleteDataMap to Badger

pzhao5:


AaronVasquez:
Looks like there's a code climate issue to fix

PR #452, opened 2018-06-27
claim_unused_prls QA completed live on the eth network

rfornea:
-adds more grifts which I used to QA the claim_unused_prls task
-QA'd to verify that brokernode successfully goes through this sequence:
      1.  Send small amount of gas to the transaction address
      2.  Use that gas to retrieve the PRL at the transaction address to the broker's wallet 
      3.  If there's some ETH still left in the transaction address, retrieve it to the broker's wallet
-adds more methods needed to accomplish the sequence
-adds unit tests of the new methods
-modifies SendETH method to be used with any "from" address that we have the key for
-adds missing down migration files

### DIFF
--- a/grifts/db.go
+++ b/grifts/db.go
@@ -7,12 +7,15 @@ import (
 	"github.com/markbates/grift/grift"
 	"github.com/oysterprotocol/brokernode/models"
 	"github.com/oysterprotocol/brokernode/services"
+	"github.com/oysterprotocol/brokernode/utils"
 	"math/big"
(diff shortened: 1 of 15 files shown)


jfarago:
@rfornea what do you think of the code climate report? Anything worth fixing?

rfornea:
@jfarago   
No.  Not at this time, anyway.  None of it is critical, and this functionality not being on the broker blocks other work from proceeding.

PR #455, opened 2018-06-29
Claim_treasure_for_webnode

rfornea:
-adds a method CheckClaimClock to get the initial claim clock status and we can later check it again and compare with previous result to see if our claim succeeded
-adds claim_treasure_for_webnode task
-adds webnode_treasure_claims model
-migration files for webnode_treasure_claims
-small fixes/refactors to other files
-adds a RAVEN_ENABLED env var so someone with a dev broker can turn their logging off even if not in unit tests
-adds some more grifts that I used to QA claim_treasure_for_webnode

I've triggered successful claim()s from the Golang code a number of times.  I posted an example of one in the weekly-updates channel.  

Still TODO for this task (PR was getting very big, so I'll add these in my next PR):
-unit tests for CheckClaimClock
-unit tests for webnode_treasure_claims
-unit tests for claim_treasure_for_webnode
-QA with the actual webnode

Up next:
-add a broker_broker_transactions job and model, so if alpha's initial sending of PRLs to beta fails it can try again, and so beta can monitor whether alpha has sent it the PRL or not and eventually punish it in the DRS if it didn't.


rfornea:
I looked through the codeclimate stuff for this PR.  None of it seems worth addressing to me, at least not with this PR.  There are some similar blocks of code, but that's because the functionality is similar but with some small critical differences that would make a refactor tedious and time-consuming.  Which is not to say that we shouldn't do it, but rather let's spend our development time on more critical things first such as:  (1) all eth/prl functionality (2) the badger and streaming changes that will support larger file sizes and (3) webnode beta testing (blocked by (1)).   

I created #457 for us to come back and get the codeclimate reports to a better state once the more critical things are in.

jobs/process_unassigned_chunks.go
```diff
@@ -283,6 +283,7 @@ func InsertTreasureChunks(chunks []models.DataMap, treasureChunks []models.DataM
 			treasureChunksInserted++
 			i++ // skip an iteration
 		} else if _, ok := treasureChunksMapped[chunks[i].ChunkIdx+idxTarget]; ok {
+			// LOOK INTO THIS
```
AaronVasquez: ?

rfornea: Oops.  This code looked suspicious to me so I made a note to take a closer look at it but forgot to remove the note.  Will remove it now.

PR #458, opened 2018-06-29
Bash script to auto deploy to staging env

EdmundMai:


travis/deploy.sh
```diff
@@ -0,0 +1,28 @@
...
+EOM
+
+TEST_BROKERS=("52.14.218.135" "18.217.133.146")
+
+for ip_address in "${TEST_BROKERS[@]}"
+do
+  ssh -o StrictHostKeyChecking=no ubuntu@$ip_address -i ./travis/id_rsa << END
+    $DEPLOY_SCRIPT
+END
+done
```
rfornea: @automyr  Would you mind having a look at this PR, since you have dealt with similar issues?

automyr: Checking.

automyr: Not sure if we have another script that sets the env vars for the broker, but if not we should include it here. For docker clean-up, we could also use the script you found yesterday just to be sure there are no remaining volumes. Everything else looks fine to me.

travis/deploy.sh
```diff
@@ -0,0 +1,30 @@
...
+  git stash;
+  git clean -f -d;
+  git checkout master;
+  git pull;
+  cp /home/ubuntu/database.dev.yml ./database.yml;
+  cp /home/ubuntu/docker-compose.dev.yml ./docker-compose.yml;
+  docker stop $(docker ps -aq);
+  docker rm $(docker ps -aq);
+  docker rmi $(docker images -q);
+  sudo curl https://gist.githubusercontent.com/mlebkowski/471d2731176fb11e81aa/raw/4a6a18a19ac5a9f1b8c6533a07f93e01da8ddba0/cleanup-docker.sh | bash;
```
rfornea: I'm going to go ahead and approve this, but we should consider whether we want to do 
`cp .env.test .env` before we do:
`sudo sed -i -e 's/127.0.0.1/$real_ip/' ./.env`

There's not, afaik, an obvious way to tell the dev brokers when we have added a new variable option to the .env vars, so the only good way I could think of to do it would be to check in some default value to .env.test and that'd get copied over.  

Also not sure, but this:
`sudo sed -i -e 's/127.0.0.1/$real_ip/' ./.env`
might need to be this:
`sudo sed -i -e 's/127.0.0.1/$real_ip/' .env`

PR #461, opened 2018-06-30
Add LogIfValidationError

pzhao5:


pzhao5:
codeclimate is bit misleading here. Please ignore it.

PR #488, opened 2018-07-11
Pre deploy qa

rfornea:
Small things that came up while QAing

### DIFF
--- a/grifts/db.go
+++ b/grifts/db.go
@@ -203,7 +203,7 @@ var _ = grift.Namespace("db", func() {
 
 		treasuresToBury := []models.Treasure{}
 
-		err := models.DB.RawQuery("SELECT * from treasures ORDER BY created_at asc").All(&treasuresToBury)
+		err := models.DB.RawQuery("SELECT * from treasures ORDER BY created_at ASC").All(&treasuresToBury)
 
 		if err == nil {
 			for _, treasureToBury := range treasuresToBury {
--- a/grifts/db.go
+++ b/grifts/db.go
@@ -203,17 +203,18 @@ var _ = grift.Namespace("db", func() {
 
 		treasuresToBury := []models.Treasure{}
 
-		err := models.DB.RawQuery("SELECT * from treasures").All(&treasuresToBury)
+		err := models.DB.RawQuery("SELECT * from treasures ORDER BY created_at asc").All(&treasuresToBury)
 
 		if err == nil {
 			for _, treasureToBury := range treasuresToBury {
 				fmt.Println("________________________________________________________")
-				fmt.Println("ETH Address:  " + treasureToBury.ETHAddr)
-				fmt.Println("ETH Key:      " + treasureToBury.ETHKey)
-				fmt.Println("Iota Address: " + treasureToBury.Address)
-				fmt.Println("Iota Message: " + treasureToBury.Message)
-				fmt.Println("PRL Status:   " + models.PRLStatusMap[treasureToBury.PRLStatus])
-				fmt.Println("PRL Amount:   " + treasureToBury.PRLAmount)
+				fmt.Println("Genesis Hash:  " + treasureToBury.GenesisHash)
+				fmt.Println("ETH Address:   " + treasureToBury.ETHAddr)
+				fmt.Println("ETH Key:       " + treasureToBury.ETHKey)
+				fmt.Println("Iota Address:  " + treasureToBury.Address)
+				fmt.Println("Iota Message:  " + treasureToBury.Message)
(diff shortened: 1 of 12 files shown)


grifts/db.go
```diff
@@ -203,17 +203,18 @@ var _ = grift.Namespace("db", func() {
 
 		treasuresToBury := []models.Treasure{}
 
-		err := models.DB.RawQuery("SELECT * from treasures").All(&treasuresToBury)
+		err := models.DB.RawQuery("SELECT * from treasures ORDER BY created_at asc").All(&treasuresToBury)
```
pzhao5: nit: ASC

rfornea: Okay.  I'll go ahead and do all of them.

services/iota_wrappers.go
```diff
@@ -458,35 +491,41 @@ func verifyChunksMatchRecord(chunks []models.DataMap, checkTrunkAndBranch bool)
...
-	} else if len(response.Hashes) == 0 {
-		filteredChunks.NotAttached = chunks
 	}
+	return matchesTangle, notAttached, doesNotMatch
+}
 
-	if len(filteredChunks.MatchesTangle) > 0 {
-		oyster_utils.LogToSegment("iota_wrappers: chunks_matched_tangle", analytics.NewProperties().
-			Set("num_chunks", len(filteredChunks.MatchesTangle)))
+func assignToChunksArray(matchFound bool, chunk models.DataMap, matchesTangle *[]models.DataMap, doesNotMatch *[]models.DataMap) {
```
pzhao5: this method does not really add much value. Consider inline

services/iota_wrappers.go
```diff
@@ -458,35 +491,41 @@ func verifyChunksMatchRecord(chunks []models.DataMap, checkTrunkAndBranch bool)
...
+			break
+		}
 	}
-	return filteredChunks, err
+	return matchFound
+}
+
+func makeTransactionObjects(transactionObjects []giota.Transaction) (transactionObjectsMap map[giota.Address][]giota.Transaction) {
+	for _, txObject := range transactionObjects {
+		transactionObjectsMap[txObject.Address] = append(transactionObjectsMap[txObject.Address], txObject)
```
pzhao5: I don't think you create this object, and it will lead to panic. See: https://play.golang.org/p/A9JbB_Q499q

services/iota_wrappers.go
rfornea: This was a codeclimate complaint.  I can change it back.

jobs/process_unassigned_chunks_test.go
```diff
@@ -321,6 +321,9 @@ func (suite *JobsSuite) Test_InsertTreasureChunks_BetaSession() {
 
 func (suite *JobsSuite) Test_SkipVerificationOfFirstChunks_Beta() {
 
+	oyster_utils.SetBrokerMode(oyster_utils.ProdMode)
```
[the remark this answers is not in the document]
rfornea: Yeah, it is only for the one case.  It resets it back to whatever is in .env when that case finishes.

grifts/db.go
```diff
@@ -243,6 +244,17 @@ var _ = grift.Namespace("db", func() {
 		return nil
 	})
 
+	grift.Desc("delete_old_genesis_hashes", "Delete all old stored genesis hashes")
+	grift.Add("delete_old_genesis_hashes", func(c *grift.Context) error {
+
+		duration := -5 * time.Hour
+
+		models.DB.RawQuery("DELETE FROM stored_genesis_hashes WHERE created_at <= ?",
```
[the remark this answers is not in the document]
rfornea: okay I'll change it.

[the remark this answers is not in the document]
rfornea: It works but TIMESTAMPDIFF seems better.

services/iota_wrappers.go
rfornea: Good catch, thanks!

services/iota_wrappers.go
```diff
@@ -437,18 +437,51 @@ func verifyChunksMatchRecord(chunks []models.DataMap, checkTrunkAndBranch bool)
...
+			break
+		}
 
-		for _, txObject := range trytesArray.Trytes {
-			transactionObjects[txObject.Address] = append(transactionObjects[txObject.Address], txObject)
+		trytesArray, err := api.GetTrytes(hashes[i:end])
+
+		if err != nil {
+			oyster_utils.LogIfError(err, nil)
+			return matchesTangle, notAttached, doesNotMatch
```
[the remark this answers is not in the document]
rfornea: Was trying to avoid doing any further logic, if there wasn't anything in trytesArray.Trytes.  But I'll just add an explicit check for a length on trytesArray.Trytes.

PR #492, opened 2018-07-12
add sleep to see if we can get migrate to work

rfornea:
Hoping we can get the buffalo db migrate call in docker-compose.yml to work.


rfornea:
Nevermind

PR #493, opened 2018-07-13
Pre deploy qa

rfornea:
-fix and/or rename some migration files that were being stubborn
-more grifts
-small QA fixes


migrations/20180621213450_add_msg_id_to_deprecate_msg.up.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE `treasures`
+ADD msg_id VARCHAR (255) DEFAULT NULL;
+
+ALTER TABLE `completed_data_maps`
+ADD msg_id VARCHAR (255) DEFAULT NULL;
+
+ALTER TABLE `data_maps`
+ADD msg_id VARCHAR (255) NOT NULL;
```
rfornea: I do not know why but for some mysterious reason the .fizz version of this file gave me errors when running migrations on the QA broker.  Would do 'migrate down' then 'migrate up' and would constantly get errors about duplicate columns.  Switched it to .sql style and it worked for me locally, will see what Travis thinks.   

For the other migrations files that I renamed, it mysteriously just didn't run them, until I changed the timestamps.  Weeeeeird.

PR #496, opened 2018-07-13
Pre deploy qa

rfornea:
Still having issues with this column.  This tries to separate it out into different columns to see if that helps.


rfornea:
Don't know what is going on with these migrations.  
Just going to deploy as-is.  I can always run SQL commands to add/change a column.

PR #509, opened 2018-07-22
Optimize the DB query

pzhao5:
Using UPDATE with WHERE will lock the entire table and prevent any further operation.
UPDATE with WHERE and SELECT will lead to error from DB.

As a result, break the query into 2 parts. SELECT and then UPDATE

### DIFF
--- a/models/upload_sessions.go
+++ b/models/upload_sessions.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math"
 	"math/big"
+	"strings"
 	"time"
 
 	"github.com/gobuffalo/pop"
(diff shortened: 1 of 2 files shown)


models/upload_sessions.go
```diff
@@ -312,29 +313,56 @@ func (u *UploadSession) GetTreasureIndexes() ([]int, error) {
...
+		}
 		err = DB.RawQuery("UPDATE data_maps SET status = ? "+
-			"WHERE genesis_hash = ? AND status = ? AND message != ? AND msg_status = ?",
+			"WHERE id IN(?)",
 			Unassigned,
-			u.GenesisHash,
-			Pending,
-			DataMap{}.Message,
-			MsgStatusUnmigrated).All(&[]DataMap{})
+			idsString).All(&[]DataMap{})
```
rfornea: I have had terrible luck being able to use IN with Pop.  https://github.com/gobuffalo/pop#using-in-queries   

I was talking with someone in the buffalo slack and they said when you pass in something for IN, it has to be an interface.  `someInterface := make([]interface{}, len(whatever))`

And then in the query in Pop you'd use it like this `someInterface...`  

Not sure if this is helpful.  I never got it to work.

pzhao5: I am going to give this up.
Since the IN() could be very big. And SQL will complain about again. I think if we lock the table, I think it will be fine

pzhao5: And it is not easy to break the IN() into batches and add more complexity. Just trust the SQL db to do the work

PR #511, opened 2018-07-25
Async

rfornea:
Changes so that MakeTreasureIdxMap can be async.  

How it was handled before:
-wait for all data map entries to be built
-make the treasure map and encrypt the eth keys all in one step in actions/upload_sessions

How it is handled now:
-go ahead and create the treasure idx map regardless of whether the chunks are ready or not.  Keys will not yet be encrypted
-in process_paid_sessions, check for sessions whose treasure keys need encrypting.  Encrypt them if the treasure chunks exist.  Otherwise, wait until the next time the task is called.

### DIFF
--- a/models/upload_sessions.go
+++ b/models/upload_sessions.go
@@ -68,6 +68,8 @@ const (
 	TreasureInDataMapComplete
 )
 
+const MaxTimesToCheckForAllChunks = 1000
+
 var StoragePeg = decimal.NewFromFloat(float64(64)) // GB per year per PRL; TODO: query smart contract for real storage peg
 
 // String is not required by pop and may be deleted
@@ -190,6 +192,10 @@ func (u *UploadSession) StartUploadSession() (vErr *validate.Errors, err error) 
 	go func() {
 		vErr, err = BuildDataMaps(u.GenesisHash, u.NumChunks)
 	}()
+	// TODO:  Remove this.  This is just to help make these changes in
+	// fewer PRs.  We may need to use WaitForAllChunks in some of our tests.
+	_, err = u.WaitForAllChunks()
+	oyster_utils.LogIfError(err, nil)
 	return
 }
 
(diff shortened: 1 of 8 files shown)


models/upload_sessions.go
```diff
@@ -187,7 +190,13 @@ func (u *UploadSession) StartUploadSession() (vErr *validate.Errors, err error)
 		DB.ValidateAndUpdate(u)
 	}
 
-	vErr, err = BuildDataMaps(u.GenesisHash, u.NumChunks)
+	go func() {
+		vErr, err = BuildDataMaps(u.GenesisHash, u.NumChunks)
+	}()
+	// TODO:  Remove this.  This is just to help make these changes in
+	// fewer PRs.  We may need to use WaitForAllChunks in some of our tests.
+	_, err = u.WaitForAllChunks(MaxTimesToCheckForAllChunks)
```
AaronVasquez: This is still blocking. Shouldn't this also go in the goroutine?

models/upload_sessions.go
```diff
@@ -68,6 +68,9 @@ const (
 	TreasureInDataMapComplete
 )
 
+/*MaxTimesToCheckForAllChunks is the maximum number of times we will check for the chunks in WaitForAllChunks*/
+const MaxTimesToCheckForAllChunks = 1000
```
AaronVasquez: We should probably set this to a really high number initially just to be paranoid and to make sure we can handle big files.

models/upload_sessions.go
rfornea: It is intentionally blocking, for now.  This is to be able to handle the various async changes in smaller PRs.

rfornea: This will be pulled out in another PR later today or tomorrow.

models/upload_sessions.go
rfornea: Sure, I can set it to like 5000 or something.

PR #528, opened 2018-07-29
Improve the purge completed datamap

pzhao5:


rfornea:
We could probably just move a data_map row with a status of "completed" or "confirmed" to completed_data_maps and remove it from data_maps, and not even worry about checking if it's one of the completed genesis hashes.  We'd know a session was complete by whether there are any rows still left in data_maps that match the session's genesis hash.

PR #539, opened 2018-08-03
Integrates with AWS Lambda

AaronVasquez:


services/iota_wrappers.go
```diff
@@ -102,7 +128,7 @@ func init() {
 		panic("Invalid IRI host: Check the .env file for HOST_IP")
 	}
 
-	provider := "http://" + host_ip + ":14265"
```
pzhao5: this will fail. revert

services/iota_wrappers.go
```diff
@@ -352,15 +379,59 @@ func sendChunksToChannel(chunks []models.DataMap, channel *models.ChunkChannel)
 		models.DB.ValidateAndSave(&chunk)
 	}
 
-	channel.EstReadyTime = SetEstimatedReadyTime(Channel[channel.ChannelID], len(chunks))
-	models.DB.ValidateAndSave(channel)
+	if os.Getenv("ENABLE_LAMBDA") == "true" {
+		go func() {
+			// Batch chunks by limit
```
pzhao5: move this to a different function

services/iota_wrappers.go
```diff
@@ -41,8 +44,7 @@ type IotaService struct {
 	FindTransactions
 	GetTransactionsToApprove
 }
-
-type ProcessingFrequency struct {
+type ProcessingFhttpRequency struct {
```
rfornea: Is `ProcessingFhttpRequency` a spelling error?

AaronVasquez: yeah, didnt mean to change that. Good catch

services/iota_wrappers.go
```diff
@@ -352,15 +383,19 @@ func sendChunksToChannel(chunks []models.DataMap, channel *models.ChunkChannel)
 		models.DB.ValidateAndSave(&chunk)
 	}
 
-	channel.EstReadyTime = SetEstimatedReadyTime(Channel[channel.ChannelID], len(chunks))
-	models.DB.ValidateAndSave(channel)
+	if os.Getenv("ENABLE_LAMBDA") == "true" {
+		go batchPowOnLambda(chunks)
+	} else {
```
rfornea: The ProcessUnassignedChunks looks at the number of ready channels and that influences how many chunks it sends.  It also calls a method (in this file, I think) that tracks how long the broker is taking to complete PoW, which determines how often process_unassigned_chunks runs.  These are things we'll probably want to change, before we deploy the lambda changes to prod.  

(This comment isn't in reference to this particular PR, just a general comment)

PR #550, opened 2018-08-16
Lambda qa

AaronVasquez:


pzhao5:
Please fix test

services/iota_wrappers.go
```diff
@@ -687,6 +694,17 @@ func lambdaWorker(provider string, lChan <-chan []*lambdaChunk) {
 		}
 		defer res.Body.Close()
 
+		fmt.Println("=========RESPONSE START=======")
+		bodyBytes, err := ioutil.ReadAll(res.Body)
+		bodyString := string(bodyBytes)
+
+		if err != nil {
+			fmt.Println(err.Error)
```
pzhao5: oyster_util.LogIfError(err, nil)

services/awsgateway/lambda.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/iotaledger/giota"
+)
+
+const (
+	// Public
+	// https://docs.aws.amazon.com/lambda/latest/dg/limits.html
+	// 6MB payload, 300 sec execution time, 1000 concurrent exectutions.
+	// Limit to 1000 POSTs and 20 chunks per request.
+	MaxConcurrency = 1000
+	MaxChunksLen   = 3000 // 3 MB
```
AaronVasquez: @rfornea No matter what I change this to, the amount of chunks returned from the query in `process_unassigned_chunks` seems to max out to 300 chunks. Any ideas why?

rfornea: It's because of the batch size.  I believe this is configured in the job_runner file.

Dockerfile
```diff
@@ -42,7 +42,7 @@ RUN apt-get install -y -q netcat
 RUN mkdir -p $GOPATH/src/github.com/oysterprotocol/brokernode
 WORKDIR $GOPATH/src/github.com/oysterprotocol/brokernode
 
-RUN go get -u github.com/gobuffalo/buffalo/buffalo
+RUN go get -u -v github.com/gobuffalo/buffalo/buffalo
```
rfornea: Was this intentional?  I'm not disagreeing, just checking.

services/awsgateway/lambda.go
```diff
@@ -0,0 +1,70 @@
...
+		Payload:      payload,
+	})
+
+	return err
+
+	// fmt.Println("=========RESPONSE START=======")
+	// fmt.Println("LAMBDA RETURNED")
+	// bodyString := string(res.Payload)
+	// fmt.Println(bodyString)
+	// fmt.Println("========= RESPONSE END =======")
```
rfornea: meant to leave these?

Dockerfile
AaronVasquez: Yeah, it just prints more stuff so we can see what's going on.

services/awsgateway/lambda.go
AaronVasquez: Yeah, it will make debugging easier if there are problems.

PR #560, opened 2018-08-22
prod/dev env for lambda

AaronVasquez:


AaronVasquez:
Hm, yeah, good idea. I'll do that in a future PR.

PR #561, opened 2018-08-23
Data maps from MySQL ---> Badger

rfornea:
No rush on approving this, take your time.  It's  a ton of changes and there wasn't really an easy way to break it up into smaller changes since it affects pretty much the entire backend.  

- store all chunk data in badger instead of MySQL
- puts all these changes behind a feature flag so we can toggle it on and off
- update all unit tests
- no longer rely on a "Status" property in each data_map row.  Instead we now will be using the following additions to the upload_sessions to know what to do with the data maps:
      - NextIdxToAttach - the broker tracks the next chunk idx it needs to attach for a session
      - NextIdxToVerify - the next chunk idx for a session that needs to be verified
      - AllDataReady - a boolean that will be true when all hashes are done, all messages are uploaded, and the treasure payload has been created
- move kv_store to utils to avoid an import cycle


rfornea:
Still need to work on the changes to allow concurrent map reads/writes, but that can wait for another PR since we've recently done a deploy.

pzhao5:
Let me know when it is ready

PR #565, opened 2018-08-30
Increase lambda concurrency

AaronVasquez:


rfornea:
Go ethereum updated their param names again.  Change params.Shannon to params.GWei and this should pass.

PR #568, opened 2018-09-05
Fixes potential race condition

AaronVasquez:
Completes the check-lock-check pattern. Avoids instantiating the singleton more than once.


rfornea:
Failing due to some ethereum related thing, probably an issue with the testnet.  Going to go ahead and merge.

PR #570, opened 2018-09-06
WIP Prioritize meta

AaronVasquez:


AaronVasquez:
@rfornea this is ready to be reviewed.

Note: For simplicity, I do PoW for the meta chunk, but dont update the `NextIdxToAttach`. It's just 1 chunk, so I don't think the performance gain outweighs the added complexity. We can optimize this after `rev2`, which would change where the meta chunk is.

jobs/check_alpha_payments.go
```diff
@@ -46,6 +47,19 @@ func CheckPaymentToAlpha() {
...
+			go func() {
+				metaChunk, err := models.GetMetaChunk(brokerTx.GenesisHash)
+				oyster_utils.LogIfError(err, nil)
+				if err != nil {
+					return
+				}
+
+				err = services.PowAndBroadcast([]oyster_utils.ChunkData{metaChunk})
+				oyster_utils.LogIfError(err, nil)
+			}()
```
rfornea: I'm not sure this will still work while we have OYSTER_PAYS set to "true"?  I don't think the for loop that this is in will ever get hit, because I think this:
```
	brokerTxs, _ := models.GetTransactionsBySessionTypesAndPaymentStatuses([]int{},
		[]models.PaymentStatus{models.BrokerTxAlphaPaymentPending})
```

...will not find any brokerTxs as long as we are in prod mode with OYSTER_PAYS set to true.

AaronVasquez: Where do you think the best place to attach meta is if OYSTER_PAYS is true?

rfornea: There's not really any obvious places in the existing code that come to mind.  Could perhaps check in ProcessAndStoreChunkData if OYSTER_PAYS is true and if the metadata chunk is one of the chunks being passed into the method, and if so, do the PoW with that chunk immediately?  

So there's basically two things we care about, whether the user has paid, and whether we have all the important data for the metadata chunk (hash and message).  The best fit might be the check_if_all_data_is_ready task, but then you'd be waiting for all the chunk data to arrive and the treasure chunk to be created to, which is potentially longer than we'd want to wait.  

Unfortunately I think the best thing to do is create a new task that runs every few seconds...it could check for (paid) alpha sessions with a next_idx_to_attach of 0 or (paid) beta sessions with a next_idx_to_attach of `session.NumChunks - 1`.  We could either add a new status to upload_sessions for when the meta is already attached, or we could use `VerifyChunkMessagesMatchRecord` or `FindTransactions` and see if anything is on the tangle for the metachunk(s) we found, and only attach if we found nothing.  

There may be a simpler option that I'm missing.

PR #574, opened 2018-09-12
Hex panic

rfornea:
I believe these changes should finally fix the hex panic issues.  I think the reason we were having these problems in production but not in the dev environment is because the aurora DB may have been truncating some of the values that we were using as keys/nonces for encryption, which would obviously make it impossible to decrypt successfully.  So now for the encryption/decryption methods we will explicitly retrieve whatever is in the db rather than just assuming the model we are working with is accurate, since the db version may be slightly different.


rfornea:
While testing this, I gave a test aurora db to one of my test brokers and was finally able to do multiple uploads without any of those hex errors.

PR #585, opened 2018-09-27
Update lambda.go

rfornea:
It works like a dream in prod when you set it to 300


rfornea:
Fixes the `Request too long` errors.

rfornea:
That's probably what happened.

PR #598, opened 2018-10-09
Update actions_test.go

rfornea:
Add `Test_ActionSuite` to v2


rfornea:
Closing this.  This approach will not work due to an import cycle.

PR #603, opened 2018-10-11
Remove some dead code and setup create method upload_session for S3

pzhao5:


actions/v3/upload_sessions.go
```diff
@@ -43,71 +46,69 @@ type UploadSessionCreateResV3 struct {
...
+		NumChunksLimit = v
+	}
+}
+
 // Update uploads a chunk associated with an upload session.
 func (usr *UploadSessionResourceV3) Update(c buffalo.Context) error {
-	start := PrometheusWrapper.TimeNow()
-	defer PrometheusWrapper.HistogramSeconds(PrometheusWrapper.HistogramUploadSessionResourceUpdate, start)
+	return c.Render(202, actions_utils.Render.JSON(map[string]bool{"success": true}))
+}
```
rfornea: Is this intended to be just a stand-in for now?

actions/v3/upload_sessions.go
```diff
@@ -43,71 +46,69 @@ type UploadSessionCreateResV3 struct {
...
+			c.Error(400, err)
+			return err
+		}
+
+		reqBetaBody := bytes.NewBuffer(betaReq)
 
-	if oyster_utils.DataMapStorageMode == oyster_utils.DataMapsInBadger {
-		dbID := []string{oyster_utils.InProgressDir, uploadSession.GenesisHash, oyster_utils.MessageDir}
+		// Should we be hardcoding the port?
+		betaURL := req.BetaIP + ":3000/api/v2/upload-sessions/beta"
```
rfornea: change to v3?

actions/v3/upload_sessions.go
```diff
@@ -43,71 +46,69 @@ type UploadSessionCreateResV3 struct {
...
-		models.ProcessAndStoreChunkData(req.Chunks, uploadSession.GenesisHash, treasureIdxMap,
-			models.DataMapsTimeToLive)
-	}()
-
-	return c.Render(202, actions_utils.Render.JSON(map[string]bool{"success": true}))
+	res := uploadSessionCreateResV3{
+		BatchSize:   BatchSize,
+		UploadAlpha: bucketRequest{},
+		UploadBeta:  uploadBeta,
+	}
```
rfornea: I may be mistaken but I think we had started leaning towards the client still uploading to the broker, and the broker itself passing the chunks on to S3.  @EdmundMai @AaronVasquez

pzhao5: But if the bucket is public-read-write, then why can't client upload directly to the bucket rather than going via the brokernode?

actions/v3/upload_sessions.go
pzhao5: Sorry, it is still in progress.

pzhao5:
Still in progress, will ping you once it is ready.

actions/v3/upload_sessions.go
rfornea: Sorry, I did not notice the "WIP"

actions/v3/upload_sessions.go
rfornea: The client could upload directly to S3, but then the client would have to know how to interact with S3.  The brokernode is the one who decides whether it will use badger or S3 and the client shouldn't need to know.  

There are tradeoffs either way.  We can discuss this more.

actions/v3/upload_sessions.go
pzhao5: Lol, updated the title

actions/v3/upload_sessions.go
pzhao5: Yes, let's do that in slack for it

AaronVasquez: We decided on having the client uploading to the broker, and then the broker uploading to s3 for a few reasons:

- Easier to implement, and we could launch the feature sooner
- Keeps the client "dumb" so it doesn't need to know which data store the broker is using
- Makes it easier to handle auth to s3 since we can store creds on the broker. The buckets will be private and only accessible by the broker and lambda

pzhao5: I am planning to update it in the next PR

PR #604, opened 2018-10-12
Extends upload_session table to include ApiVerison and S3BucketName.

pzhao5:


models/upload_sessions.go
```diff
@@ -232,6 +235,11 @@ func (u *UploadSession) BeforeCreate(tx *pop.Connection) error {
 		u.AllDataReady = AllDataNotReady
 	}
 
+	// Treat the API via api/v2 endpoint
+	if u.ApiVersion == 0 {
+		u.ApiVersion = 2
+	}
+
```
rfornea: It might be best for the status to specifically state whether the data is stored in S3 or badger, or perhaps just add another status that specifies this.  The API that we use for the upload doesn't matter so much.  Even from the v3 endpoints, it's not a guarantee that the broker will use S3 for its storage.  An individual broker can be configured to use badger or S3, and at some point we might even have logic for the broker to make a decision to use either kind of storage based on the size of the upload, how much space is left on the broker, etc.

pzhao5: So, instead of ApiVersion, do you want me to rename it as status?  I think status is kind of vague. But I am okay with it.

rfornea: It doesn't have to be called "status".  It could be called something like "StorageMethod" or "StorageStatus" or something else.  Just something to indicate where we put the data, whether it's badger or S3 or possibly something else eventually.

PR #605, opened 2018-10-16
Update the S3 API util metho

pzhao5:


actions/v3/s3_utils.go
```diff
@@ -10,27 +14,32 @@ import (
...
+		bucketPrefix = "unknown"
+	}
+	bucketPrefix = strings.Replace(bucketPrefix, "_", "-", -1)
 }
 
 /* Create unique bucket name. */
 func createUniqueBucketName() string {
-	// TODO: Figure out how to make this unique
-	return ""
+	atomic.AddUint64(&counter, 1)
```
rfornea: Is this adding the time to the unique bucket name?  This might be an issue, since only the broker who started the upload would know the name.  This could be an issue once we implement the early download logic and once we allow public brokers.  

We might need the bucket name to be something that any broker (or even the client) could derive.

rfornea: @AaronVasquez thoughts?

AaronVasquez: Would the genesis hash work?

rfornea: I think so.  As long as it's something that any broker or the client could figure out from the handle.  

Are there any risks to this?  We will make the bucket read-only once the broker is done uploading, right?

AaronVasquez: Yes, and only the brokers uploading will be authorized to write to it. This is something we might need to coordinate between alpha and beta brokers.

rfornea: Did we decide if they'd use the same bucket or different buckets?  I think they might need to use different buckets, because in the future one of the brokers could be malicious.

rfornea: Unless we have some way for the "good broker" to do something about the malicious broker.

AaronVasquez: Hm, we went with same bucket for simplicity, but you're right. One broker can be malicious.

We also went with the client not being aware of s3 at all, so the client wouldn't need to know the bucket. However, the client will then need to remember which broker it used to upload the file if it wants to access the data from s3 instead of the tangle.

rfornea: Perhaps we could still use the genesis hash and have them upload to a shared bucket. Once they start meeting in the middle on S3, they continue through the rest of the chunks checking S3 to make sure the other broker didn't upload garbage. If a broker detects that the other broker has uploaded garbage, it reports it to the DRS. Once enough brokers report the malicious broker, it doesn't get selected anymore. The only problem with this is the malicious broker would have the chance to ruin a few uploads, and there wouldn't be a way for the good broker to salvage the upload.   

If we want them to use separate buckets, the client could add a field to the metadata with the addresses of the brokers it used, or maybe the upload session IDs, or something. Then derive the bucket from some hash of genesis hash + broker address or genesis hash + session ID. But this still allows a malicious broker because each broker would know the others address.   

Maybe we will just have to store the broker addresses in the metadata and go ahead and have the brokers make unique buckets unknown to anyone else. While the upload is still in progress, the client can check the metadata for the broker address and ask it for the bucket, since the broker will still have that bucket name in it's upload_sessions table. We'd have to make sure the metadata gets attached to the tangle very quickly since the client would need the data in the metadata to get the bucket.

pzhao5: The other thought on this is that, we create 1 bucket per brokernode and make the object key deterministic.

PR #625, opened 2018-10-25
Avoid dynamic creating bucketName. Use BUCKET_NAME from env file

pzhao5:


.env.test
```diff
@@ -90,3 +90,4 @@ LAMBDA_ENV="dev"
 AWS_ACCESS_KEY_ID=""
 AWS_SECRET_ACCESS_KEY=""
 AWS_REGION=us-east-2
+BUCKET_NAME="unknown"
```
AaronVasquez: Can we name this `AWS_BUCKET_NAME` so we can kinda namespace it to be part of our AWS env vars?

PR #627, opened 2018-10-25
include numChunksLimit in status response

rfornea:
-adds numChunksLimit to status response
-return -1 if there is no limit


rfornea:
Do you guys want the URL for this to be `brokerAddress:3000/status` instead of `brokerAddress:3000/api/v2/status`?  I can make it work on either URL until we update the client to point to just `/status`.

CKH4:
`brokerAddress:3000/status` would make more sense because it won't be subject to api versions right?

rfornea:
Yeah I think it would, have updated it.

Issue #631, opened 2018-10-30
Update the latest version for iota lib


pzhao5:
We need to update the current iota lib to the latest one.

https://github.com/iotaledger/iota.go
This library has just submitted a merge request and lead to a few new changes. See here:
https://github.com/iotaledger/iota.go/commit/77d57410156f031853103f96fd3299c7594d27fd

Rather than import just "github.com/iotaledger/giota", 
We need to:
All giota.Transaction moves to here and must import this: "github.com/iotaledger/iota.go/transaction"

All giota.Address rename to trinary.Hash and import this: "github.com/iotaledger/iota.go/trinary"
All giota.ToTrytes() will be trinary.NewTrytes()
All giota.ToAddress() will be trinary.NewTrytes()

All giota.GetTransactionsToApproveResponse renamed to api.TransactionsToApprove

Use transaction.AsTransactionObjects() to convert Trytes to Transaction object.

Use FindTransactionsQuery instead FindTransactionsRequest

MrRedPandabear:
@pzhao5 At least Zhao remains cool af.

MrRedPandabear:
@MrRedPandabear Oh well.. didn't mean to do that. Sorry. 😅

PR #632, opened 2018-10-31
Fix build break for brokernode with iota lib upgrade

pzhao5:
Addressed this issue: https://github.com/oysterprotocol/brokernode/issues/631

### DIFF
--- a/jobs/process_unassigned_chunks.go
+++ b/jobs/process_unassigned_chunks.go
@@ -3,11 +3,11 @@ package jobs
 import (
 	"errors"
 	"fmt"
-	"github.com/iotaledger/giota"
 	"math"
 	"os"
 	"time"
 
+	"github.com/iotaledger/iota.go/trinary"
 	"github.com/oysterprotocol/brokernode/models"
 	"github.com/oysterprotocol/brokernode/services"
 	"github.com/oysterprotocol/brokernode/utils"
@@ -384,8 +384,8 @@ func HandleTreasureChunks(chunks []oyster_utils.ChunkData, session models.Upload
 
 	for i := 0; i < len(chunks); i++ {
 		if _, ok := treasureMap[int(chunks[i].Idx)]; ok {
-			address := make([]giota.Address, 0, 1)
-			chunkAddress, err := giota.ToAddress(chunks[i].Address)
+			address := make([]trinary.Hash, 0, 1)
+			chunkAddress, err := trinary.NewTrytes(chunks[i].Address)
 			if err != nil {
 				oyster_utils.LogIfError(errors.New(err.Error()+" in HandleTreasureChunks() in "+
 					"process_unassigned_chunks"), nil)
--- a/services/iota_wrappers.go
+++ b/services/iota_wrappers.go
@@ -10,7 +10,12 @@ import (
 	"sync"
 	"time"
 
-	"github.com/iotaledger/giota"
+	giota "github.com/iotaledger/iota.go/api"
+	"github.com/iotaledger/iota.go/bundle"
+	"github.com/iotaledger/iota.go/consts"
+	"github.com/iotaledger/iota.go/pow"
+	"github.com/iotaledger/iota.go/transaction"
+	"github.com/iotaledger/iota.go/trinary"
 	"github.com/joho/godotenv"
 	"github.com/oysterprotocol/brokernode/models"
 	"github.com/oysterprotocol/brokernode/services/awsgateway"
@@ -64,7 +69,7 @@ type VerifyChunksMatchRecord func([]oyster_utils.ChunkData, bool) (filteredChunk
 
 /*ChunksMatch defines the type for a function which returns a boolean if chunk tangle data matches chunk db data.
 This type used for mocking.*/
-type ChunksMatch func(giota.Transaction, oyster_utils.ChunkData, bool) bool
+type ChunksMatch func(transaction.Transaction, oyster_utils.ChunkData, bool) bool
 
 /*SendChunksToLambda defines the type for a function which will send the chunkds to AWS Lambda.
 This type used for mocking.*/
@@ -76,11 +81,11 @@ type VerifyTreasure func([]string) (verify bool, err error)
 
 /*FindTransactions defines the type for a function which will attempt to find transactions on the iota tangle based on
 addresses.  This type used for mocking.*/
-type FindTransactions func([]giota.Address) (map[giota.Address][]giota.Transaction, error)
+type FindTransactions func([]trinary.Hash) (map[trinary.Hash][]transaction.Transaction, error)
 
 /*GetTransactionsToApprove defines the type for a function which will get a trunk and branch transaction for proof of
 work.  This type used for mocking.*/
-type GetTransactionsToApprove func() (*giota.GetTransactionsToApproveResponse, error)
+type GetTransactionsToApprove func() (*giota.TransactionsToApprove, error)
 
 /*DoPoW defines the type for a function that does proof of work.*/
 type DoPoW func(chunks []oyster_utils.ChunkData) error
@@ -92,8 +97,6 @@ type FilteredChunk struct {
 }
 
 const (
-	maxTimestampTrytes = "MMMMMMMMM"
-
 	// MaxNumberOfAddressPerFindTransactionRequest is a hard limit
 	// on the request for FindTransaction to IOTA
 	MaxNumberOfAddressPerFindTransactionRequest = 1000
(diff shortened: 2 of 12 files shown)


services/iota_wrappers.go
```diff
@@ -299,48 +309,51 @@ func GetProcessingFrequency() float64 {
 }
 
 // Finds Transactions with a list of addresses. Result in a map from Address to a list of Transactions
-func findTransactions(addresses []giota.Address) (map[giota.Address][]giota.Transaction, error) {
+func findTransactions(addresses []trinary.Hash) (map[trinary.Hash][]transaction.Transaction, error) {
```
rfornea: Out of curiosity, is there a `Trytes` or `Address` data type?

services/iota_wrappers.go
```diff
@@ -299,48 +309,51 @@ func GetProcessingFrequency() float64 {
...
 
 	numOfBatchRequest := int(math.Ceil(float64(len(addresses)) / float64(MaxNumberOfAddressPerFindTransactionRequest)))
 
 	remainder := len(addresses)
 	for i := 0; i < numOfBatchRequest; i++ {
 		lower := i * MaxNumberOfAddressPerFindTransactionRequest
 		upper := i*MaxNumberOfAddressPerFindTransactionRequest + int(math.Min(float64(remainder), MaxNumberOfAddressPerFindTransactionRequest))
-		req := giota.FindTransactionsRequest{
+		req := giota.FindTransactionsQuery{
 			Addresses: addresses[lower:upper],
```
rfornea: It's somewhat surprising that `[]trinary.Hash` is an acceptable type to pass in to this.

jobs/process_unassigned_chunks.go
```diff
@@ -384,8 +384,8 @@ func HandleTreasureChunks(chunks []oyster_utils.ChunkData, session models.Upload
 
 	for i := 0; i < len(chunks); i++ {
 		if _, ok := treasureMap[int(chunks[i].Idx)]; ok {
-			address := make([]giota.Address, 0, 1)
-			chunkAddress, err := giota.ToAddress(chunks[i].Address)
+			address := make([]trinary.Hash, 0, 1)
```
rfornea: This might be another place to look at using a different type, if applicable.  

The reason is because iota treats hashes and addresses as different things.  Both are ultimately 81 character tryte strings, but an address is where a transaction is located on the tangle, while a hash is derived from the overall transaction trytes.  So every transaction has an address and a hash, but they are different for each transaction.  

It doesn't matter that much since both are 81 character tryte strings, and it's possible that iota has decided that hashes and addresses are so similar to just use the same type for both.  We should check.

services/iota_wrappers.go
```diff
@@ -139,11 +142,18 @@ func init() {
...
+		LocalPowFunc: bestPow,
+	})
+	if err != nil {
+		panic(err)
+	}
 
 	seed = "OYSTERPRLOYSTERPRLOYSTERPRLOYSTERPRLOYSTERPRLOYSTERPRLOYSTERPRLOYSTERPRLOYSTERPRL"
 
-	powName, bestPow = giota.GetBestPoW()
+	powName, bestPow = pow.GetBestPoW()
```
rfornea: This is not an error that you made, but we could fix it in this PR anyway.  This line should probably be moved up above line 145, before the api is instantiated.

services/iota_wrappers.go
```diff
@@ -353,30 +366,33 @@ func doPowAndBroadcast(branch giota.Trytes, trunk giota.Trytes, depth int64,
 			trytes[i].BranchTransaction = trunk
 		}
 
-		timestamp := giota.Int2Trits(time.Now().UnixNano()/1000000, giota.TimestampTrinarySize).Trytes()
-		trytes[i].AttachmentTimestamp = timestamp
-		trytes[i].AttachmentTimestampLowerBound = ""
-		trytes[i].AttachmentTimestampUpperBound = maxTimestampTrytes
+		trytes[i].AttachmentTimestamp = time.Now().UnixNano() / 1000000
```
rfornea: Do they no longer want it in trits?

services/iota_wrappers.go
```diff
@@ -455,10 +471,10 @@ func verifyChunkMessagesMatchRecord(chunks []oyster_utils.ChunkData) (filteredCh
 func verifyChunksMatchRecord(chunks []oyster_utils.ChunkData, checkTrunkAndBranch bool) (filteredChunks FilteredChunk, err error) {
 
 	filteredChunks = FilteredChunk{}
-	addresses := make([]giota.Address, 0, len(chunks))
+	addresses := make([]trinary.Hash, 0, len(chunks))
```
rfornea: here

services/iota_wrappers.go
```diff
@@ -505,10 +520,10 @@ func verifyChunksMatchRecord(chunks []oyster_utils.ChunkData, checkTrunkAndBranc
 	return filteredChunks, err
 }
 
-func filterChunks(hashes []giota.Trytes, chunks []oyster_utils.ChunkData, checkTrunkAndBranch bool) (matchesTangle []oyster_utils.ChunkData,
+func filterChunks(hashes []trinary.Trytes, chunks []oyster_utils.ChunkData, checkTrunkAndBranch bool) (matchesTangle []oyster_utils.ChunkData,
 	notAttached []oyster_utils.ChunkData, doesNotMatch []oyster_utils.ChunkData) {
 
-	transactionObjectsMap := make(map[giota.Address][]giota.Transaction)
+	transactionObjectsMap := make(map[trinary.Hash][]transaction.Transaction)
```
rfornea: here

services/iota_wrappers.go
```diff
@@ -569,15 +587,15 @@ func checkTxObjectsForMatch(transactionObjectsArray []giota.Transaction, chunk o
 	return matchFound
 }
 
-func makeTransactionObjects(transactionObjects []giota.Transaction) (transactionObjectsMap map[giota.Address][]giota.Transaction) {
-	transactionObjectsMap = make(map[giota.Address][]giota.Transaction)
+func makeTransactionObjectsMap(transactionObjects []transaction.Transaction) (transactionObjectsMap map[trinary.Hash][]transaction.Transaction) {
+	transactionObjectsMap = make(map[trinary.Hash][]transaction.Transaction)
```
rfornea: here

services/iota_wrappers.go
```diff
@@ -599,10 +617,10 @@ func chunksMatch(chunkOnTangle giota.Transaction, chunkOnRecord oyster_utils.Chu
 // Verify PoW of work.
 func verifyTreasure(addr []string) (bool, error) {
 
-	iotaAddr := make([]giota.Address, len(addr))
+	iotaAddr := make([]trinary.Hash, len(addr))
```
rfornea: here

utils/segment_logger.go
```diff
@@ -48,9 +50,9 @@ func GetLocalIP() string {
 	return "cannot get local ip"
 }
 
-func MapTransactionsToAddrs(txs []giota.Transaction) (addrs []giota.Address) {
+func MapTransactionsToAddrs(txs []transaction.Transaction) (addrs []trinary.Hash) {
 
-	addrs = make([]giota.Address, 0, len(txs))
+	addrs = make([]trinary.Hash, 0, len(txs))
```
rfornea: here and 2 lines above

services/iota_wrappers.go
pzhao5: There is not more Address data type. And it is replaced by Hash
Hash is Trytes.

services/iota_wrappers.go
pzhao5: I think Address is backed by Hash(and by Trytes), so it makes sense to remove the Address type and just use Hash

services/iota_wrappers.go
pzhao5: Good catch!

utils/segment_logger.go
pzhao5: Sorry, this comment is not very clear to me. Can you explain?

jobs/process_unassigned_chunks.go
pzhao5: Correct,

See here:
https://github.com/iotaledger/iota.go/blob/5ef833c00afc210d70375c2f25e86be6a9ff6b9b/transaction/transaction.go#L22
Both Hash and Address has the same type

services/iota_wrappers.go
rfornea: Okay, we don't have to worry about it in this PR.  But in a PR in the future, we should replace most uses of `trinary.Hash` with `trinary.Trytes`.  All hashes are tryte strings, but not all tryte strings are hashes, if that makes sense.  Everything will still work either way.

services/iota_wrappers.go
pzhao5: It is now in int64

services/iota_wrappers.go
pzhao5: It is not clear to me about this comment

jobs/process_unassigned_chunks.go
rfornea: Okay sounds good.  Disregard the trytes/address/hash comments.

utils/segment_logger.go
rfornea: I was pointing out places where we should perhaps use the `Trytes` type instead of `Hash`, but disregard.  I did not know at the time that iota itself had decided to use the Hash type for addresses.

PR #633, opened 2018-11-02
Add DeleteObjects API to batch delete a set of files on buckets

pzhao5:


rfornea (approved):

rfornea (commented):
still a few iota issues it seems.

PR #634, opened 2018-11-02
Update Create/CreateBeta method

pzhao5:
Saving the following data into S3 so that lambda could process it.

1. Data format as gensis_hash/data/file_index. File contains a BatchSize of data.
2. Genesis Hash as gensis_hash/hash/file_index.
3. Each data and genesis hash is 1:1 mapping. And there is not treasure within.
4. Config data(single file) so that lambda knows how to process it.
5. Treasure map data (single file).

When lambda attach to the data, it would first download config data and treasure map.
And then start to upload each data/genesis hash in pairs.
When arrive to a particular index that need to skip (for inserting the treasure), Lambda will then insert it with treasure rather than the data.

### DIFF
--- a/actions/v3/upload_sessions.go
+++ b/actions/v3/upload_sessions.go
@@ -292,9 +292,9 @@ func saveTreasureMapForS3(u *models.UploadSession, treasureIndexA []int, treasur
 		u.MakeTreasureIdxMap(mergedIndexes, privateKeys)
 
 		// Verify that MakeTreasureIdxMap is correct. Otherwise, regenerate it again.
-		treasureIndexes, _ := alphaSession.GetTreasureIndexes()
-		if alphaSession.TreasureStatus == models.TreasureInDataMapPending &&
-			alphaSession.TreasureIdxMap.Valid && alphaSession.TreasureIdxMap.String != "" &&
+		treasureIndexes, _ := u.GetTreasureIndexes()
+		if u.TreasureStatus == models.TreasureInDataMapPending &&
+			u.TreasureIdxMap.Valid && u.TreasureIdxMap.String != "" &&
 			len(treasureIndexes) == len(mergedIndexes) {
 			break
 		}
--- a/actions/v3/upload_sessions.go
+++ b/actions/v3/upload_sessions.go
@@ -56,7 +56,7 @@ type uploadSessionConfig struct {
 	BatchSize        int    `json:"batchSize"`        // Represent each data contain not more number of field.
 	FileSizeBytes    uint64 `json:"fileSizeBytes"`    // Represent the total file size.
 	NumChunks        int    `json:"numChunks"`        // Represent total number of chunks.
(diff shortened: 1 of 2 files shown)


actions/v3/upload_sessions.go
```diff
@@ -44,9 +45,18 @@ type uploadSessionCreateBetaResV3 struct {
...
+	BatchSize     int            `json:"batchSize"`
+	Invoice       models.Invoice `json:"invoice"`
+}
+
+/*uploadSessionConfig represents the general configuration that other client could understand.*/
+type uploadSessionConfig struct {
+	BatchSize        int    `json:"batchSize"`        // Represent each data contain not more number of field.
+	FileSizeBytes    uint64 `json:"fileSizeBytes"`    // Represent the total file size.
+	NumChunks        int    `json:"numChunks"`        // Represent total number of chunks.
+	ReserveIteration bool   `json:"reserveIteration"` // Represent whether iterate it from the beginning to end or end to the beginning.
```
rfornea: nit:  ReverseIteration

actions/v3/upload_sessions.go
```diff
@@ -219,3 +264,51 @@ func sendBetaWithUploadRequest(req uploadSessionCreateReqV3) (uploadSessionCreat
...
+	for {
+		privateKeys, err := EthWrapper.GenerateKeys(len(mergedIndexes))
+		if err != nil {
+			return oyster_utils.LogIfError(errors.New("Could not generate eth keys: "+err.Error()), nil)
+		}
+		if len(mergedIndexes) != len(privateKeys) {
+			return oyster_utils.LogIfError(errors.New("privateKeys and mergedIndexes should have the same length"), nil)
+		}
+		// Update treasureId
+		u.MakeTreasureIdxMap(mergedIndexes, privateKeys)
```
rfornea, line 292: The v2 version of this file has a few extra lines after this line.  Starts here and ends on line 228:
https://github.com/oysterprotocol/brokernode/blob/master/actions/v2/upload_sessions.go#L219  

They may not all be essential.  I feel like there was a reason I had to add all those, but it's been so long I don't remember what the reason was.

utils/utils.go
```diff
@@ -391,3 +388,37 @@ func IntMax(x, y int) int {
...
+
+/*GetObjectKeyForTreasure will return the treasure key.*/
+func GetObjectKeyForTreasure(genesisHash string) string {
+	return fmt.Sprintf("%v/%v", genesisHash, "treasure_map")
+}
+
+/*GetObjectKeyForData will return object key for particular data. startIndex is the smallest index in data. */
+func GetObjectKeyForData(genesisHash string, startIndex int, totalCount int, isReserveIteration bool, batchSize int) string {
+	index := 0
+	if isReserveIteration {
```
rfornea, line 405: nit isReverseIteration

rfornea (dismissed):

rfornea:
Went ahead and approved this but we should see if we can figure out what the deal is with Travis

rfornea (approved):

